PR #73, opened 2021-05-15
Image support

st3v3nmw:
Fixes #34.
Fixes #25.

PR #94, opened 2021-05-20
store original front and back text to put back after review

rubbish:
This I think fixes the issue I brought up in https://github.com/st3v3nmw/obsidian-spaced-repetition/pull/73#issuecomment-845484017

I'm not sure where else the original back and front text could be used. I don't think Cloze cards will run into this same issue.


st3v3nmw:
Looks good, thanks for the fix!

PR #210, opened 2021-07-27
Allow flashcard height and width to be customizable

rubbish:
I needed to be able to change this so that on iPad it's prettier and doesn't blow up images to fill up the whole screen. 

I did remove the largeDisplayMode setting as that's now doable by just making width and height 100%. That is a breaking change, but I'm not sure how much it was actually used. It's an easy fix on the user's side to just adjust both to 100%. Thoughts?


st3v3nmw:
Yes, I agree. That setting was not as flexible and useful to most users.

Issue #203, opened 2021-07-22
[FEAT] Multiple review queues


erichalldev:
Thank you so much for this plugin! I mainly use the note reviewing functionality. It would be great if separate queues could be maintained, kind of like having separate decks. For example one for "#workideas" and "#personalideas". That way I could review them separately.

dummifiedme:
+1
A choice on reviews by tags would be amazing. It could act like decks for review :D

erichalldev:
I've made a first pass at implementing this. There's a little more to do, but the bones are there, I think. https://github.com/erichalldev/obsidian-spaced-repetition/commit/069619c56b40879ef3d80e2e2bf855a7a34aef9f

st3v3nmw:
Hey @erichalldev, looks good so far!

erichalldev:
Thanks @st3v3nmw. It doesn't look like I'll have a lot of time to work on it in the next month or two, unfortunately.

st3v3nmw:
@erichalldev, if you have a minute or two, you could open a PR to the `multiple-note-review-queues` branch that I've created so that I can pick up from where you've left. Thanks.

erichalldev:
Thanks for creating that branch for me, @st3v3nmw! I just created PR #234. I wish I could spend more time on this because it's been such a game-changer for my workflow and I want to give back as much as I can.

dummifiedme:
@erichalldev, @st3v3nmw 
Does it work now? and how to enable it?

st3v3nmw:
@dummifiedme, not yet. Should be available in the next release once I finish up.

st3v3nmw:
Released!

erichalldev:
This is great! Thank you so much.

PR #234, opened 2021-08-09
Merge WIP for multiple review queues

erichalldev:
See issue #203 for context

I didn't have as much time to spend on this as I wanted so I can't guarantee correct behavior everywhere. But I believe it's a good starting point.


st3v3nmw:
Thanks for taking the time to work on this!

PR #266, opened 2021-08-23
Switch to esbuild

TfTHacker:
So this is an opinionated suggestion, but I recommend switch to esbuild  over rollup. Rollup takes on my machine about 5 seconds to compile, esbuild about 4 milliseconds. This is useful for testing.

I am using this with my two plugins and its working really well, so I suggest giving esbuild an install and try it for yourself.

This outputs to a build folder and is excluded in .gitignore

Feel free to DM me on Discord if you want to discuss further.


st3v3nmw:
This runs much faster for me as well. & the build size is much smaller.

Could you remove the main.js file from the diff so that I can merge this? Thanks.

TfTHacker:
Let me close the request, and resubmit. Sorry this was a mistake for sure on my part.

Issue #269, opened 2021-08-24
Editing cards during the review


myrnaw:
### Discussed in https://github.com/st3v3nmw/obsidian-spaced-repetition/discussions/267

<div type='discussions-op-text'>

<sup>Originally posted by **myrnaw** August 24, 2021</sup>
Hi,
Thank you for making this awesome plugin!
I know we can jump to the file by clicking the link that shows in the card, and then edit it. But it's more smooth editing during the review.
Is it possible to support this feature?</div>

TfTHacker:
I agree this would be useful. However, the approach this plugin uses is to use a modal for review. So nothing else can be edited while this is open.  To  implement this would require some difficult changes.

I submitted a PR to change the "Open  file"  to "Edit Later". It opens the current flashcard in the editor in the background.  It will  allow you to keep testing, but to have any files you clicked on "edit later" to be waiting for you after the review.

st3v3nmw:
Duplicate of #433
I'm cleaning up the issues list and that other issue has more discussion

PR #280, opened 2021-08-29
Change "Open file"  to "Edit later"

TfTHacker:
Make the "open file" open the file in the background for editing after the SR session ends. Thus, move the user to the next flashcard for review

This would  be an alternative solution  for the issue: https://github.com/st3v3nmw/obsidian-spaced-repetition/issues/269


st3v3nmw:
Nice, just in time for the next release

PR #292, opened 2021-09-11
Using tags from User Preferences and not their full name that includes the sibling

benhughes:
This fixes the issue of defineing a parent tag as a review tag and the plugin using the full tag which can clutter things up dramatically. e.g.

in the setting I have  #reference as a tag to review 

Screenshot 2021-09-11 at 22 08 22

but in the UI the plugin uses all the sibling tags like so 

Screenshot 2021-09-11 at 22 07 41

This fix uses the tags from the user preference in the UI to get this:

Screenshot 2021-09-11 at 22 58 25

Let me know your thoughts.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -290,15 +290,14 @@ export default class SRPlugin extends Plugin {
             const tags = getAllTags(fileCachedData) || [];
 
             let shouldIgnore = true;
-            for (const tag of tags) {
-                if (
-                    this.data.settings.tagsToReview.some(
-                        (tagToReview) => tag === tagToReview || tag.startsWith(tagToReview + "/")
-                    )
-                ) {
-                    if (!Object.prototype.hasOwnProperty.call(this.reviewDecks, tag)) {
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -290,15 +290,14 @@ export default class SRPlugin extends Plugin {
...
             let shouldIgnore = true;
-            for (const tag of tags) {
-                if (
-                    this.data.settings.tagsToReview.some(
-                        (tagToReview) => tag === tagToReview || tag.startsWith(tagToReview + "/")
-                    )
-                ) {
-                    if (!Object.prototype.hasOwnProperty.call(this.reviewDecks, tag)) {
-                        this.reviewDecks[tag] = new ReviewDeck(tag);
+            const matchedNoteTags = [];
```
benhughes, line 293: added this to store the matched tags otherwise I would had to repeat the logic later in the function.

st3v3nmw (approved):

Issue #289, opened 2021-09-07
[BUG] webm doesn't work 


superouioui:
**Describe the bug**
Unable to listen to webm recordings.

! [[Recording 20210826175024.webm]]

**To Reproduce**
- Recording a sound with the core plugin 'Audio recorder'. 
- Made a flashcard with the balise ![[yourSound.webm]]
- Test in preview
- That works
- Test in flashcard 
- that doesn't work

**Screenshots**
Edit:
<img width="502" alt="edit" src="https://user-images.githubusercontent.com/90142988/132397052-ef3f8dd2-5ff7-4c7e-b3aa-eb4ee8eaa629.PNG">
Preview:
<img width="612" alt="preview" src="https://user-images.githubusercontent.com/90142988/132397079-e4fd94d1-7ddf-4d4e-8a62-876c09880e0f.PNG">
Flashcard:
<img width="613" alt="flashcard" src="https://user-images.githubusercontent.com/90142988/132397093-37fd80df-767c-4b63-9bee-be4f84934c35.PNG">

**Versions (please complete the following information):**
OS: Windows 10 / Android 10
Obsidian version: v0.12.15
Plugin version: v1.6.2

superouioui:
same problem with mp3

GollyTicker:
As a frequent user of obsidian-spaced-repetition, I also +1 this, as I would highly appriciate if generally meida embeddings and possibly iframes work in flashcards the same way how they do in the normal preview.

I am not sure however, whether this is really a bug. After all, the [feature set](https://github.com/st3v3nmw/obsidian-spaced-repetition#features) only says that images and latex etc work - but not media in general.

If this behavior currently is expected, then this issue might be an enhancement.

careilly:
Definitely an enhancement: the code seems to handle only IMG embeds. (Actually, treats any embed that isn’t a markdown file as an image.)

careilly:
I have mp3s working on my local version, I’ll workout webm and how to do a pull request. 

There must be a more general purpose embedding api that could be used here, but that’s not a quick fix.

Avinash-Bhat:
looks like the release for this is not yet done. @st3v3nmw any ETA on that?

st3v3nmw:
Hi all, audio and video now work in the new release (v1.7.0). Thanks to @careilly's work!

PR #297, opened 2021-09-18
Updated to display mp3 and webm audio on flashcards

careilly:
Fixes PR #289 to display mp3 and webm audio so they can be played


st3v3nmw:
Hey @careilly, thanks for this!

PR #309, opened 2021-10-11
Add fullscreen mode

rgruenewald:
This change adds a new toggle option called "Fullscreen Mode" to the settings, which basically sets the min-height and min-width css property to 100%. The result is that the modal is always using 100% of the screen size, regardless of the actual card content. 

- On my iPad and iPhone it removes the ~20% size top margin -> 20% more card content visible!
- On my desktop it hides any note content in the background, which might disturb from the actual content. -> increased focus!


st3v3nmw:
Closing as stale

PR #315, opened 2021-10-31
Changing interpolate to an optional arg of t

LouisStAmour:
While we can't change the existing Obsidian API that sometimes extends string with new methods, it's not necessary to extend string when we only need interpolation when calling t for translated strings...

This PR makes it a small bit easier to test scheduling.ts by avoiding the need to provide a global interpolate function for every string.

Provided some unit tests for translation and interpolation, just to make sure it continued to work as expected.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -3,6 +3,7 @@ module.exports = {
     preset: "ts-jest",
     testEnvironment: "jsdom",
     moduleNameMapper: {
-        "src/(.*)": "<rootDir>/src/$1"
-    }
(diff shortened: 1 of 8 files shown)


LouisStAmour:
Also note that I didn't update how we get locale - right now, if locale happens to change, it won't be reflected in calls to t unless the entire import is reset and reloaded/re-executed from scratch. That's why the tests for the translate function look a bit messy, it's a sign that there's a small bit of tech debt here.

Issue #216, opened 2021-08-01
[BUG] Note transclusion in flashcards doesn't work


OliverBalfour:
**Describe the bug**
If I use a transclusion in a flashcard like `![[question file]] :: answer` it just renders as the file name in plain text.

**Expected behavior**
I would expect the file's contents to be loaded in place of the link

**Versions (please complete the following information):**
 - OS: Manjaro Linux
 - Obsidian version: v0.12.12
 - Plugin version: v1.5.7

**Additional context**
I realise the note review feature is designed for this kind of thing, but it'd be convenient to be able to include small files in a flashcard prompt using the Obsidian syntax.

crazymedguy:
Seconding this!!! Need it badly. 

**Here's why this can be gold, and can create an entirely new method of SR in Obsidian:**

In the Cornell notetaking method, you have a column of cues next to the actual notes. The cue column does not modify or touch the actual notes at all. So you do not have awkward syntax within the notes with A?B or A:B. You also do not have questions followed by answers, which is too conversational and Socratic in nature. Very weird to have them in notes. This is why the cues live on their own, *pointing* to the notes. 

In the same way, we can create a flashcard document which can serve as a "cue column" that does not modify the original note documents, and then set the answer as a ![[transclusion]] to some other place in the text. 

This can ultimately work as "programmable attention". Just like when we receive Notification alerts on Twitter/Email/push notifications on mobile, our attention can be programmed and shifted around. In other words, we are not bound to having to create A::B or A?B syntax. This "overlay cue column" serves as a way to store SR information, and to point attention to relevant parts of the knowledge graph.

TfTHacker:
This would be useful for sure, basically interconnecting thoughts from other notes into a flashcard.

we recommend support for transclusion of headings and block refs

Sachin-12:
I too would love this feature. I like to have the questions seperate like in Cornell system and embed the answers for each question.

cjl12377:
Really looking forward to this feature. Any idea how high on the priority list it is?

st3v3nmw:
The new release (v1.7.0) supports note transclusion for entire notes & heading-level. Thanks to @joelmeyerhamme's work.
Block references pending...

ronzulu:
Hi @OliverBalfour 

It appears that this had been resolved a fair while ago, so I'll close this now.

Feel free to reopen if this is still an issue.

Cheers
Ronny

PR #325, opened 2021-11-25
transclusions initial implementation

joelmeyerhamme:
After grumbling a bit about the Obsidian API, I managed to implement support for transclusions in flash cards. I am not familiar with Typescript or the toolchain, but this worked well for me. It can definitely do with some cleanup, is there something else I need to do? (implements #216)


st3v3nmw:
Hey @joelmeyerhamme, thanks for this!

Issue #349, opened 2021-12-29
[BUG] HTML comments from other plugins cause the creation of some cards to be left out.


andrewcrook:
HTML comments from other plugins cause the creation of some cards to be left out.

**example** 
IDs for 3 cards from the plugin Obsidian to Anki 

<img width="884" alt="image" src="https://user-images.githubusercontent.com/803618/147673884-a7e78850-a960-4de2-a7d7-63627ecb56c4.png">

**result** 
2 cards created for spaced repitition the middle one missing 

** expected **
3 cards for spaced repetition 

** cause **
code checks for the closing of a HTML comment on the next line when it should be the same line.

https://github.com/st3v3nmw/obsidian-spaced-repetition/blob/5514b1a2159bcade9071f3661b2cd66017a2f0f6/src/parser.ts#L38

andrewcrook:
Pull request created #350

st3v3nmw:
PR merged, released in v1.7.0

PR #350, opened 2021-12-29
[Bug fix] for HTML Comment parsing

andrewcrook:
as seen in #349 
code checking for the comment end checks on the next line rather than the same line 
This causes some causing parsing issues.


st3v3nmw:
Hi @andrewcrook, thanks for the fix!

PR #402, opened 2022-02-13
fix: handle code fences better

zcysxy:
Markdown has two code fences syntaxes: multiple backticks and multiple tildes. It also supports nested code fences. For example

`````markdown
An admonition:

````ad-note

```git
+ print("hello")
- print("world")
```

~~~python
print("hello world")
~~~
````
`````

This pull request makes the plugin support the more general Markdown code fences syntax.


st3v3nmw:
Thanks!

PR #409, opened 2022-03-01
Add shortcut to the review button text

carlosbaraza:


st3v3nmw:
I'm closing this since we don't wanna show extra information that's only needed by power users. My assumption is that power users will look at the docs and learn the keyboard shortcuts and non-power users can use as-is. Just like in Anki :).
Thanks for the contribution though!

PR #411, opened 2022-03-04
Add a lock file

olingern:
👋 Awesome project. 

I didn't see a `yarn.lock` or `package-lock.json` so thought I would open a PR. Yarn still appears to be considerably faster than npm, so I've added a `yarn.lock`. If this looks good to you, I'll update the documentation as well

You can read more here about lock files
https://classic.yarnpkg.com/blog/2016/11/24/lockfiles-for-all/


st3v3nmw:
LGTM.
Merged, thanks!

PR #412, opened 2022-03-04
Configure jest

olingern:
I noticed jest wasn't including all covered files. There's also a small TypeScript update in this due to Jest failing on the inferred HTMLElement type

### Before
image

### After
image


st3v3nmw:
I've made these changes

PR #413, opened 2022-03-04
Add and run prettier

olingern:
It looks like prettier wasn't ran on all files. This localizes it as a devDepedency, adds a cmd, and adds a Github Action job


st3v3nmw:
Thanks! I'll be fixing the files with linting issues..

PR #416, opened 2022-03-12
Update zh-cn.ts

11-check-it-out:
1. Removed the space between Chinese word and English word according to the specification CY/T 154-2017.
2. “回应已收到”->“反馈已收到” to make the meaning clearer


st3v3nmw:
Thanks!

PR #417, opened 2022-03-12
translate korean

ksundong:
Hi, I translated it to Korean.
Thank you for developing!


st3v3nmw:
Thanks for translating!

PR #425, opened 2022-03-21
Add "ogg" as supported audio file format

sabarrett:
Was pulling my hair out trying to figure out why an audio file wouldn't embed -- turns out ogg wasn't supported! I modified the constant in the exported .js file and now it works just fine.


st3v3nmw:
Thanks!

Issue #431, opened 2022-03-31
 Custom syntax or {{Curly Brackets}} for cloze flashcards


DryIce1:
Would it possible to add either the ability to make a custom cloze syntax or add {{Curly Brackets}} cloze syntax.

I would prefer to only use ==highlights== in my notes for highlights. 

Thanks!

_Originally posted by @DryIce1 in https://github.com/st3v3nmw/obsidian-spaced-repetition/discussions/244#discussioncomment-1555405_

MotaOcimar:
Related to #387

st3v3nmw:
Fixed in #438, pending release, will update after change goes out

MotaOcimar:
in #439 XD

st3v3nmw:
> in #439 XD

Oops, updated

Issue #387, opened 2022-01-25
[FEAT] Adding support for strikethrough/custom cloze


dummifiedme:
**Is your feature request related to a problem? Please describe.**
Currently highlights are utilized which turn out to ne impotant as visual cues for me while reading. If we could have custom clozes, we could have out own symbols to work with!!

**Describe the solution you'd like*
We could have any set of symbols for this plug-in to search for. We can just enclose the cloze in those symbols which we want to assign.
eg `__cloze__`or `~~ cloze~~`or `₹₹cloze₹₹`Anything really.

AB1908:
I believe cloze symbols can be altered in settings.

dummifiedme:
> I believe cloze symbols can be altered in settings.

there are two modes, bold or highlight. Thats it.

AB1908:
Ah, you're right. I confused this with other settings.

AB1908:
I think I might be able to add this but will need time.

dummifiedme:
> 



> I think I might be able to add this but will need time.

If you cn please do. Happy to wait..

st3v3nmw:
Hi,

I've just made a release [v1.13.0](https://github.com/st3v3nmw/obsidian-spaced-repetition/releases/tag/1.13.0) that ships with custom cloze patterns 🥳 all thanks to #943. The updated documentation is live [here](https://www.stephenmwangi.com/obsidian-spaced-repetition/flashcards/cloze-cards/#cloze-types).

I've tested the release but not as extensively as I normally would. Perhaps we need a beta version of the plugin? That being said, I'm on standby to fix bugs ASAP. Just give me a shout.

Cheers.

PR #439, opened 2022-04-25
Add curly brackets as option for cloze fields

lachlancollins:
This PR adds {{curly bracket}} syntax as an optional delimiter for cloze deletions. This is a bit more similar to Anki's syntax, and also does not overlap with any Markdown syntax meaning it doesn't get styled (I want to be able to use highlighting and bolding, but not necessarily where I want to make clozes).

I've chosen {{double brackets}} over {single brackets}, but have no strong preference and can change it if that is preferred.

Related: #387 #431


adzcai:
Just wanted to bump this -- Any updates? I'd also like to see this feature!

st3v3nmw (commented):
Looks good to me, thanks!

putzwasser:
I you'd change it to single curly brackets this would be compatible with https://github.com/reuseman/flashcards-obsidian and would allow exporting the flashcards to anki.

Issue #433, opened 2022-04-02
[FEAT]Editing in real time when reviewing


Jack20152019:
**Is your feature request related to a problem? Please describe.**
one：Currently, When we review the CARDS can only edit later, but when we have reviewed all the CARDS，Prepare to edit it，Many ideas have been forgotten，even edit which card also can not find, a lot of ideas not record in time，It is very bad for me.
two:The references of  obsidian are not able to be viewed.like this :[[reference]]

**Describe the solution you'd like**
We can real time editing like ''anki''

**Additional context**

faerch90:
+ one for this feature! 

I don't know if it is possible to replicate, but in Logseq's implementation you can edit directly in the modal. 
It could be a source of inspiration for this request.

23727:
this would be great; I am looking forward to it!

AB1908:
I want this as well and am annoyed enough to consider giving it a stab but sometimes, I need to decompose a card into two or more cards. I can't think of an easy way to do this TBH.

AB1908:
See also #269

AB1908:
So I've implemented a very rudimentary version of this but had to remove the keyboard shortcuts since I couldn't figure out how to offload event handlers when there's a new modal on top. I'm going to go ahead and release that since I don't use keyboard shortcuts anyway lol.

AB1908:
Managed to add a hacky workaround to keep keyboard shortcuts enabled.

AB1908:
In my fork, I ended up rewriting a good chunk of the plugin and removing a ton of features I didn't personally use but I do have editing support. Feel free to try it out via BRAT. That said, it is primarily designed for my own use. <https://github.com/AB1908/obsidian-spaced-repetition>

Issue #414, opened 2022-03-07
[FEAT] Edit Later Note


23727:
When choosing to edit a flashcard for later, it would be helpful to add an option to write yourself a message where you can state how/why this card needs to be edited.

AB1908:
Hmm, sounds useful. How do you envision this to work?

23727:
> Hmm, sounds useful. How do you envision this to work?

RemNote is a good example,

![image](https://user-images.githubusercontent.com/84650530/157397756-050a048f-323a-4496-ac10-dbd1b7e27c82.png)

the user is given the option to edit later, where they can write a note and 'suspend' the card until further notice

the card is actually tagged with 'edit later' and can be viewed on the tag page. the user ten edits the card and removes the edit later tag by clicking 'done editing'

![image](https://user-images.githubusercontent.com/84650530/157398021-f0e118ec-9c89-4f28-80fe-dc2e9b2c3c99.png)

23727:
the purpose if this function is to indicate why we want to edit the tag, "move" it to someplace temporary, and then continue our review session. 

at the end of our review session, we can go through the cards that need to be edited and we'd have a good idea of why we chose to edit them later.

AB1908:
Hmm, in my fork, I opted to tag it. Can move around easily enough with the search pane. However, the edit modal idea is interesting but I'm not sure we can actually edit a file in a modal. If it's possible, we'd probably have to write back to the file and that can be tricky. I wonder if, for a simple test, we can create a popup modal and use that with the tag.

AB1908:
Hey I got around to implementing a slightly different version of this. Instead of a descriptive note, you can now edit the card itself in a modal on the fly. I suppose you can use that functionality to actually add descriptive text and tags yourself. Test it out if you can find the time.

KyleKlus:
Hi, so we plenty of editing options now (Edit in modal, open card in new tab), so I consider this one closed😊.

Issue #376, opened 2022-01-16
[FEAT] Mark cards for editing with a tag


AB1908:
I'm aware of #269 and a PR that was merged for opening the file in the background and I'm not particularly interested in on the fly editing. Rather, I'd like to mark certain questions with a tag to edit later, such as `#refactor`. That way, I can later jump around with tags and then bulk remove with a regex once I'm done. I imagine that this is very doable and I'd love to take a crack at it. Would you be able to give me any helpful pointers?

AB1908:
FYI, my fork has this feature. I haven't bothered cleaning it up for merging though.

PR #459, opened 2022-06-20
Add card editing modal and edit later tagging

AB1908:
To follow up on the general ideas of #414, #433, #269, #376, I went ahead and added two new features:
- An edit later tag which has a default value of `#edit-later`. Can be toggled on as a setting.
- A card editing modal, inspired by chhoumann/quickadd's `wideInputPrompt` class.

Both implementations are rather hacky IMO but I'm open to suggestions to clean it up.


AB1908:
I'd also like to split the "edit later" impl and the "edit in modal" implementation into two diff version based on commit for cleaner history purposes but not sure how to do that. Do I add tags to commits?

FelixGibson:
@AB1908 happy to do that. I have added some interesting and might useful features in my fork. I'll  try to pull request to the upstream these days.

AB1908:
Sounds great! I go by the same handle on Obsidian's Discord in case you want to chat.

st3v3nmw:
Hey @AB1908, please resolve the merge conflicts so that we push this. I'd love these changes to go in. Thanks!

AB1908:
Will do! There were a few questions:

- do you want to keep the open edit later in background option? I switched it to adding a tag.
- my flashcard counts are wrong and I couldn't figure out how to update the deckTree correctly. I think I needed to delete the flashcard.

st3v3nmw:
Thanks.

> do you want to keep the open edit later in background option? I switched it to adding a tag.

A tag is better, I think.

> * my flashcard counts are wrong and I couldn't figure out how to update the deckTree correctly. I think I needed to delete the flashcard.

Lemme look into that bit of the code & see if I can figure out what's up

AB1908:
I finally finished that today. I'll still need to pull down the changes and test locally.

st3v3nmw:
Merged via https://github.com/st3v3nmw/obsidian-spaced-repetition/pull/587, thanks!
Will be trying this out over the next few days.

AB1908:
Err, I'm guessing it probably won't work and you should probably revert this haha. I'm redesigning my fork to support crud from the ground up.

st3v3nmw:
Oops, just saw this.
I tested it a little and it works fine, built a little on top of it too on https://github.com/st3v3nmw/obsidian-spaced-repetition/pull/589. Hopefully it won't be too difficult to redesign on top of the current master branch.
Anyways, it's all good! 👍🏾

PR #467, opened 2022-07-13
[feat] add modal history feature

FelixGibson:
I find it will be much more convenient if the deck list can remember which entry the user have clicked before exiting the window. So I implemented this feature and have tested for several monthes. It works well so I do PR to the upstream.


AB1908:
Oh this is hella awesome! Nice work!

st3v3nmw (commented):
Looks good to me, thanks!

PR #480, opened 2022-07-29
style: set the flashcard's font size as user editor

edvardchen:
Currently, the font size of the flashcard is 16px, although the user might change the font size for the editor in the setting panel. 

It will be reasonable to set the flashcard's font size as the user set

### DIFF
--- a/styles.css
+++ b/styles.css
@@ -74,6 +74,7 @@
 }
 
 #sr-flashcard-view {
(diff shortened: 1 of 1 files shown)


ikuygithub:
Yeah I really want to change font size on flashcards.
It's really hard to see their hard-baked and micro sized font....

st3v3nmw:
Thanks!

PR #488, opened 2022-08-30
fix Edit Later some experience problems

mengzairanshao:
1. When clicking Edit Later on the card, if the file is already open, it will not open a new one.
2. Fixed this when clicking Edit Later on the card, which originally does not scroll to the middle of the screen, and if in Preview mode, does not scroll.


src/flashcard-modal.tsx
```diff
@@ -369,8 +402,10 @@ export class FlashcardModal extends Modal {
                 scheduling.push(currCardSched);
             }
 
-            this.currentCard.cardText = this.currentCard.cardText.replace(/<!--SR:.+-->/gm, "");
-            this.currentCard.cardText += "<!--SR:";
+            this.currentCard.cardText = this.currentCard.cardText.replace(
```
st3v3nmw, line 405: This change seems unrelated with the PR title

src/main.ts
```diff
@@ -756,6 +757,12 @@ export default class SRPlugin extends Plugin {
                         );
                     siblingMatches.push([side1, side2]);
                     siblingMatches.push([side2, side1]);
+                } else if (cardType === CardType.CallOutQuestion) {
```
st3v3nmw, line 760: This too

src/parser.ts
```diff
@@ -84,6 +94,14 @@ export function parse(
             }
             cardText += "\n" + codeBlockClose;
             i++;
+        } else if (/^(( *> *)+\[!question\]-* *)/gm.test(lines[i])) {
+            cardType = CardType.CallOutQuestion;
```
st3v3nmw, line 98: And this

st3v3nmw (changes requested):
I love the edit later fixes, we could remove the extra unrelated code or move those particular changes to a separate PR? Thanks!

st3v3nmw:
Sorry but we have to close this since we have incoming work to modify flashcards directly from the modal and we've removed the edit later functionality. Please see https://github.com/st3v3nmw/obsidian-spaced-repetition/pull/459 and https://github.com/st3v3nmw/obsidian-spaced-repetition/pull/589.
Thanks for your contribution!

PR #506, opened 2022-10-08
Bug 495 card should be added to all tagged decks not just the first one

### LINKED ISSUE #495: [BUG] Obsidian SRS ignores more than one tag
**Describe the bug**
Obsidian SRS ignores more than one tag. It seems as if the algorithm stops assigning a card/note to decks after it matched the first tag it knows.

**To Reproduce**
1. Create this flashcard:

```
---
tags:
  - flashcard
  - flashcard/test
  - flashcard/test2
  - otherFlashcardTag
---
**Spaced repetition** is an **evidence-based** learning technique that is usually performed with **flashcards**.

Newly introduced and more difficult flashcards are shown **more frequently**.

```

2. Add #`otherFlashcardTag` to the list of flashcard tags.
3. Open the review modal.
4. The card will be filed under `flashcard` only.

**Expected behavior**
The card should be IMHO filed under 
  - `flashcard`
  - `flashcard/test`
  - `flashcard/test2` and
  - `otherFlashcardTag`

I would at least expect the card to filed under `flashcard` and `otherFlashcardTag`. It would be best if it belonged to all three decks.

Use case: A flashcard for marketing that is about market research and customer behaviour.
I'd file the card under `#marketing/marketResearch` and `#marketing/customerBehaviour` and maybe `cognitiveBias`. So, I'd like to study the card if I study the `marketing/marketResearch` or the `marketing/customerBehaviour` deck or the whole `marketing` deck if got a lot of time. Furthermore I'd like to see the card if I study `cognitiveBias`es.

**Screenshots**
image
Shows the file and that it only belongs to the `flashcard` deck.
image
List of tags

-   OS: `Manjaro Linux`
-   Obsidian version: `v0.15.9`
-   Plugin version: `v1.8.0`
-   If on desktop, Installer version: `5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) obsidian/0.15.9 Chrome/100.0.4896.160 Electron/18.3.5 Safari/537.36`

MostlyArmless:
I have a change which fixes #495, allowing a card to be added simultaneously to multiple decks if the note is tagged with multiple valid deck tags.

This does lead to some questionable behavior, where if a user tags the same note with `#parentTag` and `#parentTag/childTag`, then the cards in that note will be single-counted in the `#parentTag/childTag` deck (which is fine), but double-counted in the `#parentTag` deck (which seems weird, but hey it's what they've literally asked for by doing that).

# Example of what I mean

Here's the contents of the note:
image

and here's the resulting UI from this plugin when you use the 'cram flashcards in this note' command:
image

I'm not sure if this is how we want the plugin to behave, or if we would perhaps like to make it configurable by having a setting that determines what happens in this situation. I'm open to suggestions, as I haven't been using the `obsidian-spaced-repetition` plugin personally for very long so I don't have a good feel for what makes sense here.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -590,8 +590,8 @@ export default class SRPlugin extends Plugin {
         let deckPaths: string[][] = [];
         if (this.data.settings.convertFoldersToDecks) {
             deckPaths[0] = note.path.split("/");
-            deckPaths.pop(); // remove filename
-            if (deckPaths.length === 0) {
+            deckPaths[0].pop(); // remove filename
+            if (deckPaths[0].length === 0) {
                 deckPaths[0] = ["/"];
             }
         } else {
--- a/src/main.ts
+++ b/src/main.ts
@@ -196,8 +196,8 @@ export default class SRPlugin extends Plugin {
                 const openFile: TFile | null = this.app.workspace.getActiveFile();
                 if (openFile && openFile.extension === "md") {
                     this.deckTree = new Deck("root", null);
-                    const deckPath: string[] = this.findDeckPath(openFile);
-                    await this.findFlashcardsInNote(openFile, deckPath);
+                    const deckPaths: string[][] = this.findAllDeckPaths(openFile);
+                    await this.findFlashcardsInNote(openFile, deckPaths);
                     new FlashcardModal(this.app, this).open();
                 }
             },
@@ -210,8 +210,8 @@ export default class SRPlugin extends Plugin {
                 const openFile: TFile | null = this.app.workspace.getActiveFile();
                 if (openFile && openFile.extension === "md") {
                     this.deckTree = new Deck("root", null);
-                    const deckPath: string[] = this.findDeckPath(openFile);
-                    await this.findFlashcardsInNote(openFile, deckPath, false, true);
+                    const deckPaths: string[][] = this.findAllDeckPaths(openFile);
+                    await this.findFlashcardsInNote(openFile, deckPaths, false, true);
                     new FlashcardModal(this.app, this, true).open();
                 }
             },
@@ -308,11 +308,11 @@ export default class SRPlugin extends Plugin {
                 }
             }
 
-            const deckPath: string[] = this.findDeckPath(note);
-            if (deckPath.length !== 0) {
+            const deckPaths: string[][] = this.findAllDeckPaths(note);
+            if (deckPaths.length !== 0) {
                 const flashcardsInNoteAvgEase: number = await this.findFlashcardsInNote(
                     note,
-                    deckPath
+                    deckPaths
                 );
 
                 if (flashcardsInNoteAvgEase > 0) {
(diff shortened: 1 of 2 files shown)


src/main.ts
```diff
@@ -586,34 +586,33 @@ export default class SRPlugin extends Plugin {
...
+            const tagsInThisNote = getAllTags(fileCachedData) || [];
 
-            outer: for (const tagToReview of this.data.settings.flashcardTags) {
-                for (const tag of tags) {
+            for (const tagToReview of this.data.settings.flashcardTags) {
+                for (const tag of tagsInThisNote) {
                     if (tag === tagToReview || tag.startsWith(tagToReview + "/")) {
-                        deckPath = tag.substring(1).split("/");
-                        break outer;
+                        deckPaths.push(tag.substring(1).split("/"));
```
MostlyArmless, line 604: This is the crux of the change. Just append to an array here instead of early-returning from the loop.

st3v3nmw:
> This does lead to some questionable behavior, where if a user tags the same note with `#parentTag` and `#parentTag/childTag`, then the cards in that note will be single-counted in the `#parentTag/childTag` deck (which is fine), but double-counted in the `#parentTag` deck (which seems weird, but hey it's what they've literally asked for by doing that).

Yeah, this approach would break a few things like the counting as you've mentioned and we'd have the card in multiple decks, so if we review it the first time in deck A, the same card will show up later in the same review session (in deck A/B).

We'll probably need to rewrite parts of this plugin to use hash(cardText) as IDs to make it easier to remove/pick/update the same card in multiple decks

st3v3nmw (commented):

Stephen,

pikatwinky:
Any chances to check this merge conflicts @MostlyArmless ? This PR looks really really promising...

MostlyArmless:
Sure I'll take a look tonight.

ronzulu:
Hi @4Source 

Yes, if actually been quite involved, but I'm at the tail end. I'm hoping to release a beta version in the next couple of days. Will you be able to test and give me feedback?

4Source:
I think that should be possible. I'll wait for the release. Will you mark this here when the time comes?

ronzulu:
Hi @4Source @MostlyArmless @pikatwinky (and anyone else)

I have completed the update but would appreciate beta testing and feedback before I finalise the PR.

The draft PR together with the updated code is available at:
https://github.com/st3v3nmw/obsidian-spaced-repetition/pull/834

Cheers
Ronny

ronzulu:
Hi @MostlyArmless have you had a chance to look at https://github.com/st3v3nmw/obsidian-spaced-repetition/pull/834?

After all your good work with this PR, I want to make sure you are happy with the final result.

Cheers
Ronny

ronzulu:
This has been superseded by #834, which was based on the refactored code base.

PR #510, opened 2022-10-11
Added new setting to collapse/expand subdecks

MostlyArmless:
# Description of change
Added new setting determines whether the nested subdecks will initially displayed as collapsed or expanded. Default is to expand them, which is the opposite of previous behavior. Also added localizations for the new UI text in each of the supported languages using Google Translate.

# New Settings UI
New settings UI, I put it between the `Notes` and `Algorithm` section of the settings:
spacedRepetitionNewSettingDemo

# Example of the effect on the Deck UI
e.g. if you have a note tagged with `#flashcards/psychology`, and then you run the "Cram flashcards in this note" command, the previous behavior was this:
image

and the new default behavior is this:
image


st3v3nmw:
Looks good, thanks!

Issue #370, opened 2022-01-10
[FEAT] add command "cram flashcards from all notes" and/or option to disable adding SR HTML comments


blackace72:
**Is your feature request related to a problem? Please describe.**
I like the "cram flashcards in note" option, thank you for adding it. I wish it would allow you to cram from all your notes however, because it is tedious to keep opening up "cram" for all my notes. I thought maybe I should just do a normal "review all flashcards" and then delete the HTML comments afterwards, but that is even more tedious due to possibly needing to learn RegEx, having flashcards on MANY files, etc.

**Describe the solution you'd like**
A few options:
- Allow a toggle/option in settings to disable saving the SR html comment/SR data for flashcards (hopefully it can be separate for flashcards and notes, so I can SRS notes, but cram flashcards)
- Create a command like "cram flashcards from ALL/EVERY note"

**Additional context**
Great job! Love the plugin.

pathophysiologicalchariotaleeas:
I also want this!

+1 for this👍🏻

I would like to see something like "cram from all notes"
Or "custom cram"(by selecting decks)

blackace72:
Thank you. Cramming flashcards from any and all decks works now! Thanks so much.

PR #515, opened 2022-10-20
Add per-deck cramming command

### LINKED ISSUE #418: [FEAT] Cram All Cards in Specified Deck
**Is your feature request related to a problem? Please describe.**
Currently, we have the ability to cram flashcards of a particular file but not in an entire deck

**Describe the solution you'd like**
I would like to expand on the above concept by allowing the user to cram all flashcards of a particular deck

**Additional context**
Example, instead of a user cramming Integration and then Differentiation, they would be able to cram Calculus. I imagine being able to right click the deck from the queue window and having an option to 'cram cards in deck' 

This would be incredibly helpful when notes are atomic and cramming each note individually would be a lot more cumbersome than cramming per deck

LennyPhoenix:
This builds on top of #291, and adds a command to cram flashcards from any deck.

Should resolve #418


LennyPhoenix:
Would probably also resolve #370

st3v3nmw (approved):
Thanks for the contribution!

PR #543, opened 2022-12-12
Fix back button

st3v3nmw:


richardrl:
Hey @st3v3nmw as of the current obsidian / plugin release, the back button brings us to the deck page. But what if we want to "undo" a card we just did? Like we accidentally clicked red but we meant green. Can we have that back button?

PR #569, opened 2022-12-31
Implementation of same-day reviews for 1 minute & 10 minutes.

psaunderualberta:
This PR implements the ability to review a card in 1 minute or 10 minute intervals, (partially) addressing issues #502 and #503. These minute intervals are hardcoded and so an interval of 2.5 minutes, for example, is not possible. Several major changes were required to implement such a feature. These include:
1. Changing the 'interval attribute of a flashcard' from days to minutes
2. Changing the format of the review date to include hours, minutes, & seconds
3. Tracking flashcards in a heap rather than an array, using a custom comparison function (Deck.comparator in flashcard-modal.tsx)
   1. Required addition of [heap.js](https://www.npmjs.com/package/heap) NPM package.
4.  Changing content of the "Card" interface.

There are several topics I would like cleared up before I open this PR for review & approval. One thing I'm unclear on is the use of sibling cards. I would appreciate if other maintainers could explain the point of tracking sibling cards. Specifically, what's the point of lines 811-819 in flashcard-modal.tsx:
```ts
                // look for first unscheduled sibling
                const pickedCard: Card = this.newFlashcards[pickedCardIdx];
                let idx = pickedCardIdx;
                while (idx >= 0 && pickedCard.siblings.includes(this.newFlashcards[idx])) {
                    if (!this.newFlashcards[idx].isDue) {
                        modal.currentCardIdx = idx;
                    }
                    idx--;
                }
```
I also understand that it may be too early in the roadmap to implement, as the scheduling algorithm is being changed and that might want to be merged first. My hope in making this pull request is to get this feature implemented before or soon after the winter semester starts. I greatly enjoyed using this plugin in the fall semester, but found myself falling back on Anki due to the lack of same-day reviews. Having all flashcards in one place would make my own life and the lives of other users much easier.

Still TODO before approval:
- Revert build location to build/main.js. 
- Include other languages for updated constant strings
- Implement option to not randomize card review order. Currently, program forces randomization of flashcards
- Update scheduling.test.ts to use minutes
- Testing with multiple nested decks & flashcards.
- Testing with sibling cards (what are these)
- Ensuring ability to review notes has not changed.

Let me know if anyone has questions or concerns about this PR! I'd be happy to address them.

PR #614, opened 2023-01-29
update date format to YYYY-MM-DD-HH-mm and support review before due

newle:


src/flashcard-modal.tsx
```diff
@@ -490,7 +490,9 @@ export class FlashcardModal extends Modal {
             return;
         }
 
-        const dueString: string = due.format("YYYY-MM-DD");
+        const dueString: string = due.format("YYYY-MM-DD-HH-mm");
```
st3v3nmw, line 493: Isn't the `YYYY-MM-DD HH:mm` format better and more readable?

st3v3nmw (changes requested):

Stephen,

src/main.ts
```diff
@@ -527,7 +527,7 @@ export default class SRPlugin extends Plugin {
         ease = schedObj.ease;
 
         const due = window.moment(now + interval * 24 * 3600 * 1000);
-        const dueString: string = due.format("YYYY-MM-DD");
+        const dueString: string = due.format("YYYY-MM-DD-HH-mm");
```
ronzulu, line 530: I would suggest that the time component is only included if necessary. This would keep the HTML commands in the current format, and only the longer format when needed.

ronzulu:
Hi @newle, nice idea in your PR.

There seems to be similarity in purpose between your PR and #569. Both aim to provide a mechanism so that newly learned cards can be reviewed the same day (or multiple times that day).

@st3v3nmw and I are hoping to work through the outstanding PR requests, but we might only want to incorporate either your PR or #569. Can you think of a case where your method gives a better user experience over #569?

Also I'm curious what value for `reviewBeforeDue` do you use, or do you find that you change it at times?

Thanks
Ronny

ronzulu:
Unfortunately no response from OP for 6 months, closing this now.
Feel free to reopen and continue the discussion.

PR #644, opened 2023-03-12
Text inputs for clozes

goran-zdjelar:
# Fix: issue with only being able to have one cloze per card
**Description**: before the fix, you could only put one cloze per card as it can bee seen on the gif below

cloze-not-working

# Feature: replaced [...] with input fields that are later being evalueted

**Description**: instead of showing [...] now there are input fields that you can fill and after you press `Show Answer` it will give you feedback if your input was correct or no. In case it's not correct it will show you the correct answer.

cloze-inputs

 ## Note
I moved creation of Flash Card `back` to `flashcard-modal.tsx` as the values of inputs can be evalueted only after they are rendered.


AB1908:
Wow this is actually pretty cool. I will try and steal this for my fork :)

fastrick:
Awesome idea. I think this feature can be placed as option in the config file, or using a hashtag option in the line.

mariomui:
This is super cool. but seems like it has conflicts.

st3v3nmw (changes requested):

Stephen,

KyleKlus:
This feature was implemented now in the pr #1388 😊

PR #650, opened 2023-03-17
Make reset button (flashcard review modal) enabled for both new and due cards

polyipseity:
This allows putting the flashcard back temporarily while learning it.

Currently I simply add `<!--SR:!<date of today>,1,250-->` to my flashcards whenever I learn them to enable the reset button. It would be nice to have this ability builtin instead.


st3v3nmw (dismissed):

Stephen,

polyipseity:
Done.

ronzulu:
Hi @polyipseity 

Thanks for this, I updated the PR title from "Make reset button always enabled" because the button isn't "always" enabled.

It is still disabled whilst the question is shown, but always shown when the answer is.

Cheers
Ronny

ronzulu (approved):
All good

st3v3nmw:
I was about to make a release and after testing the changes, it looks like a change in #873 reverted some of your changes (apologies for that). I've fixed that by reintroducing the missing changes. It looks like the reset button now behaves the same way as skipping the new card 🤔. Am I missing something?

I was thinking that it should put the new card at the back of the queue so that you can see it later.

Thoughts?

polyipseity:
Yes, it works as expected. Thanks for the heads up.

PR #656, opened 2023-03-28
Fix issue requiring collapseIcon to be clicked twice to expand/collapsed deck tree

### LINKED ISSUE #655: [BUG] Expanding nested decks requires two clicks the first time
**Describe the bug**
When you open the deck viewer,  and click the dropdown arrow, you must click twice to expand to show the child decks.
image

**To Reproduce**
1. Open the deck viewer modal from the button on the left bar.
2. Click the arrow to the left of the deck name

Note that this does not expand until clicked a second time.

**Expected behavior**
The deck should expand to show subdecks

-   OS: Confirmed on both Linux (Fedora + flatpak) and MacOS
-   Obsidian version: v1.1.16
-   Plugin version: 1.10
-   If on desktop, Installer version: Both from obsidian plugin store and manual installation from [releases](https://github.com/st3v3nmw/obsidian-spaced-repetition/releases/download/1.10.0/obsidian-spaced-repetition-1.10.0.zip)

**Additional context**
I have found the problem in source and will be opening a pull request ASAP. Noting the issue here for documentation only.

adam-jw-casey:
Closes #655

The problem is that the initial value of `collapsed` was being set to the `shouldBeInitiallyExpanded` config value, even though these booleans should be opposite (collapsed vs. expanded).

Because the initial UI was set from `shouldBeInitiallyExpanded`, but the callback to the click event checked `collapsed`, the first click would make no visual change, but merely set the `collapsed` variable to be correct, and the second click would work as expected.

Technically only the  change on line `846` is necessary, but the other two changes prevent the same visual confusion from occurring in the future by ensuring that that the UI and `collapsed` are in sync.

### DIFF
--- a/src/flashcard-modal.tsx
+++ b/src/flashcard-modal.tsx
@@ -843,13 +843,13 @@ export class Deck {
         );
         const shouldBeInitiallyExpanded: boolean =
             modal.plugin.data.settings.initiallyExpandAllSubdecksInTree;
-        let collapsed = shouldBeInitiallyExpanded;
+        let collapsed = !shouldBeInitiallyExpanded;
         let collapseIconEl: HTMLElement | null = null;
         if (this.subdecks.length > 0) {
             collapseIconEl = deckViewSelf.createDiv("tree-item-icon collapse-icon");
             collapseIconEl.innerHTML = COLLAPSE_ICON;
             (collapseIconEl.childNodes[0] as HTMLElement).style.transform =
-                shouldBeInitiallyExpanded ? "" : "rotate(-90deg)";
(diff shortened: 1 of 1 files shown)


src/flashcard-modal.tsx
```diff
@@ -843,13 +843,13 @@ export class Deck {
         );
         const shouldBeInitiallyExpanded: boolean =
             modal.plugin.data.settings.initiallyExpandAllSubdecksInTree;
-        let collapsed = shouldBeInitiallyExpanded;
+        let collapsed = !shouldBeInitiallyExpanded;
```
adam-jw-casey, line 846: This is the actual bugfix

Issue #698, opened 2023-06-29
[BUG] Incorrect parse the reversed card


khoinguyen:
**Describe the bug**
Both multiline reversed card and inline reverse card not work correctly at first.
After one review, the HTML comment SR tag appear after the answer, the reversed card start work correctly.

**To Reproduce**

1. Open Obsidian Sandbox Vault
2. Install this plugin
3. Create a new note with following content

```
#flashcards

# The question
??
Answer1

Question2
??
Answer2

question3:::answer
```

4. Use the command pallet to `review this note`
5. The card show question while the reverse card expected to show the answer.
6. The same card can appear again with the answer (correctly as it is reverse card)
7. Finish the deck, back to the note will see the HTML comment after each answer `<!-- SR:... -->` 
8. Review again, it will show as expected: the answer as front

**Expected behavior**
From the very first time it should show the answer as front

**Screenshots**
NA

**Versions (please complete the following information):**

-   OS: MacOS Ventura
-   Obsidian version: 1.3.5
-   Plugin version: 1.10.1
-   If on desktop, Installer version: 5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) obsidian/1.3.5 Chrome/112.0.5615.183 Electron/24.3.1 Safari/537.36

**Additional context**
NA

khoinguyen:
ok, I read the code, looks like it is by intention. 

Should we add this behavior to the doc to prevent confusion?

PR #699, opened 2023-06-30
doc: update note about the reversed card

khoinguyen:
Update the document to avoid confusion about the sibling behaviour of reversed card

See https://github.com/st3v3nmw/obsidian-spaced-repetition/issues/698


st3v3nmw:
Thanks for the contribution!

Issue #690, opened 2023-05-29
[BUG] Flashcards are broken if the separator line contains spaces


carloslab-ai:
**Describe the bug**
Many times my flashcards do not appear in the deck.
I've tried to reload Obsidian multiple times in the hope that they would appear but in the end I became desperate.

Until today I just realized it's because I accidentally added spaces to the separator line.😭
These spaces are invisible so I can not detect them.

**To Reproduce**

1. Add a new flashcard
2. Add spaces to the separator line

For example:

abc
??_ _(add spaces here)
def

3. Review this deck
4. This flashcard does not appear in the deck

**Expected behavior**
The flashcard appear in the deck regardless of whether there were spaces in the separator line or not

**Versions (please complete the following information):**

-   OS: Android
-   Obsidian version: 1.4.5
-   Plugin version: 1.10.1

AB1908:
Good report. Would you like to be the one to fix it? It should be a rather simple change in the parser.

richardrl:
Very interesting find @carlos-truong !!!! This could be the reason why my cards were also invisible. #685

richardrl:
I am working on fixing this. 

Another thing I want to simultaneously fix: we need a way to track which cards do not have ANY question separator (such as `multilineCardSeparator`.

Right now, if you make a card with NO separator, the card becomes invisible. Suppose you have 1000 cards. It becomes completely impossible to track down which cards are incomplete (i.e. no separator, you only took some rough notes on the card, you didn't clearly write down a question/answer).

My workflow with cards is I am constantly putting notes down, then adding a tag like #flashcards/my_subject. But often these notes/cards are incomplete.

But under the current system, it becomes impossible to sift through my massive number of cards to see which cards are incomplete. 

Perhaps I can make the parser also return notes, which do not have the separators, so the user knows which cards are "incomplete".

Or, we can treat cards without a separator as a new card type - i.e. more like a note, than a flashcard.

Thoughts? @AB1908 @st3v3nmw

AB1908:
Hey Richard, these are pretty good ideas but I actually had some deeper thoughts on SRS and the ideas about context. I've been working on a separate plugin to actually address that as I'm finding it hard to help with this one. I have a flow in Figma as well. I think we should chat sometime as I'd really value your feedback.

matterai:
Hey guys, 

[...]

Issue #685, opened 2023-05-18
[BUG] Missing lots of cards


richardrl:
**Describe the bug**
If I check the tag #flashcards/my_subject, I see 292 cards. If I check the GUI by clicking the flashcard app, I see 180 cards. Lots of cards are missing. Any idea how to fix?

**To Reproduce**

**Expected behavior**

**Screenshots**
If applicable, add screenshots to help explain the bug. This could be the logs in the developer console (`CTRL` + `SHIFT` + `i`).

**Versions (please complete the following information):**

-   OS: [e.g. iOS]
-   Obsidian version: [e.g. v0.12.4]
-   Plugin version: [e.g. v1.4.9]
-   If on desktop, Installer version: [run `this.navigator.appVersion` on the developer console, e.g. `5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) obsidian/0.12.3 Chrome/89.0.4389.128 Electron/12.0.6 Safari/537.36"`]

**Additional context**
Add any other context about the problem here (e.g. the markdown producing the error).

st3v3nmw:
Hmm, that's weird. I'll look into that over the weekend. Hopefully with a fix out!

st3v3nmw:
Also, how do your flashcard tags settings look like?

richardrl:
@st3v3nmw I have all default settings.

Also, just to share screenshot of the issue:
![image](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/5181990/fc46df8c-251a-42d4-a1d6-ecf467e27b0a)
(270 results)

![image](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/5181990/555a0a75-9200-47b4-9f8a-e281c0cd160b)
only 163 total cards under this tag in my flashcards app

richardrl:
Hey @st3v3nmw any update here? Anything else I can do to diagnose it?

Perhaps one test you could do to replicate it is to create a bunch of random cards (like 100) under a nested tag. You will probably see the same issue.

richardrl:
@st3v3nmw Hey any update here, have you had a chance to look into this?

richardrl:
Strangely, now I have 277 results from the search bar and 59 + 28 + 164 = 251 in the plugin GUI. 

@st3v3nmw Did you implement some type of delay or something where plugins don't show up immediately? 

The remaining 26 could be because of #690 but I'm not sure.

[...]

PR #701, opened 2023-07-03
Fixing missing cards

richardrl:
Addresses #690 and #685

Fixing #690 means the parser now accepts white space before and after the separator.

Fixing #685 required breaking changes.

I was missing hundreds of cards as described in #685  - it turns out because these cards were not perfectly formatted. It is impossible to manually search through 1000s of cards to find the few that did not have the separators applied properly.

So, I have created a new note type CardType.Note which does not require a backside. It just has an empty string. Any note which does not have a separator, will still have a card generated under this new default card type.

**Note**: My change for #685 is breaking because we can no longer have multiple multiline cards in one Obsidian note. I am arguing that this change is an important improvement in the design of OSR. Previously, we used white space to generate multiple cards within one note. This leads to extremely poorly formatted cards when you have a lot of text - imagine trying to read paragraphs of text with no line breaks!

This new design decision says that each obsidian note must correspond to one card. I believe this is the correct choice - the only other alternative I see is to require some custom "note-break" symbol like "END_CARD" - but this is tedious and harms readability of cards. Forcing each note to be one card allows us to have expressive notes, with line breaks, and does not require us to insert nasty note-break symbols throughout our notes.

I believe these changes, particularly #685 are extremely important to the usability of the app. Previously, I was seriously hindered in my studies as hundreds of notes were invisible!

Under my new system in this PR, if you mis-format a note, you can easily find it later as you are studying your queue of cards. Then, if you wish, you can re-format your note into a nice front-back Q&A format.

Here is a screenshot showing the number of cards before these changes:

image

Old version of OSR: 
image

Welcome your thoughts!
@AB1908 @st3v3nmw


ronzulu:
Hi Richard

Great that you have fixed this, but are you sure that all existing users would agree with the breaking change.

The plug-in has been downloaded over 120,000 times (amazing!). I would imagine that there would be some portion of the user base that expects the current functionality to continue.

Personally, I typically have multiple questions within the same note, albeit usually single line. Still, I don't want to have to go through all my notes to find the ones that have multiple multi-line questions, and extract into separate files.

I understand that in your workflow, you temporarily have a note with a flashcard tag, as well as the front side.

If there was a way to find all notes that have the required tag but no questions (both front/back) would that solve the issue?

Regards
Ronny

ronzulu:
Hello again

Another idea is if we essentially implement https://github.com/st3v3nmw/obsidian-spaced-repetition/issues/729.

This would mean that any note that is suitably tagged, but doesn't have any questions, would appear as:
Front - Filename
Back - Contents of the note.

Cheers
Ronny

package-lock.json
st3v3nmw, line 1: Please use `pnpm` instead

PR #719, opened 2023-08-10
Remove the spaces before and after the question marks in the multi-line basic flashcards

matterai:
Fixed issue mentioned here: #690. 

**Changes:** 
- Defined `currentLine` variable in the beginning of the main cycle in `src/parser.ts` 
- Used `trim()` to remove the extra spaces when appending `cardText`
- Used `trim()` to clean the line before comparing with separator symbol

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1,12 +1,4 @@
-// import { CardType } from "src/scheduling";
-
-export enum CardType {
-    SingleLineBasic,
-    SingleLineReversed,
-    MultiLineBasic,
-    MultiLineReversed,
-    Cloze,
-}
+import { CardType } from "src/scheduling";
 
 /**
  * Returns flashcards found in `text`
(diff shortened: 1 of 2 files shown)


src/parser.ts
```diff
@@ -27,28 +27,29 @@ export function parse(
...
+            while (i + 1 < lines.length && !currentLine.includes("-->")) i++;
             i++;
             continue;
         }
 
         if (cardText.length > 0) {
             cardText += "\n";
         }
-        cardText += lines[i];
+        cardText += currentLine.trim();
```
st3v3nmw, line 48: For the card text, we should probably do a `trimEnd()` to be on the safer side. For instance, if my card contains code that's indented we'll lose the indentation if we trim both ends

st3v3nmw (changes requested):
Hi, thanks for the fix!

st3v3nmw (approved):

Issue #730, opened 2023-08-21
[BUG] Clozures are treated as multiline flashcards


Stefanuk12:
```md
- Hello ==World==
- Foo ==Bar==
```

Ideally, each bullet point would be treated as a separate flashcard but they're both merged into one. When viewing the "World" card, "World" is hidden but "Bar" is not and vice versa.

After looking at the parser, this happens because the cards are finalised once it meets an empty new line. However, there should be a check for if the clozure is not within a multiline card and if it's not, add the card in and do not consider the next line.

Stefanuk12:
It appears this is intended behaviour; a solution has been proposed in the PR.

Stefanuk12:
Duplicate of #450

Issue #450, opened 2022-06-02
[BUG] Cloze Cards require blank lines before and after to create atomic card


timblaktu:
**Describe the bug**
The cloze cards generated by this plugin for my notes that include ==cloze highlights== *include the entire multiline section (_paragraph_) of notes which is bounded by newlines above and below.*

Therefore, to create an atomic/small/one-line cloze card from my inline notes, I have to put line breaks before and after the note for the line containing the ==highlighted cloze(s)==. (From my perspective, #147 is not really fixed, interpreting it literally.)

This "paragraph-only cloze" behavior imposes too much formatting on my notes, requiring my notes to have blank newlines in between all intended clozes, preventing things like a cloze in a middle list element. The situation does not seem to be improved by adding structural markdown formatting, like headings. **It appears that the cloze parsing in this plugin assumes a "paragraph" style close by default. I believe this is backwards, because...**

Cards for Spaced Repetition are supposed to be concise and short. Source notes used to generate the cards should be free to stay concise as well. It is for these reasons that I seek to use "highlight clozes" as my primary card type, simply annotating vital words and/or phrases in my notes with highlights, and using this plugin[^1] to review them as concise flashcards. I do not plan to use the *Basic* and *Multiline* *Q & A* style in my notes.

**To Reproduce**
1. New install of plugin, turn on "Save scheduling comment on the same line..", all else defaults
2. Put #flashcard tag in a file and add several lines with highlight clozes, including lists, *intended to be separate cards*, e.g.:
```markdown
Blah, Blah, Blah
==IPSEC== provides ==Authentication== 
  - only known, trusted ==peers== can connect
  - blah, blah, blah
```
3. **Spaced Repetition: Review flashcards from all notes**
4. Notice that the 3 cards generated all have the entire "paragraph" surrounding the lines with clozes, as shown below. 

The ***problem*** here is that ***the generated cards all include the adjacent lines.***

![image](https://user-images.githubusercontent.com/12355715/171727219-f06b4cda-7523-4083-8b38-e2f87a40cc1a.png)

**Expected behavior**
I expect each line to be processed as a separate cloze note (producing multiple cards, in Anki parlance).

In the above example, this means each generated card should **not** include the adjacent "Blah" lines.

**Versions (please complete the following information):**

-   OS: Windows 10
-   Obsidian version: v0.14.6
-   Plugin version: 1.7.2
[...]

PR #731, opened 2023-08-22
Support for atomic clozures

Stefanuk12:
# Addresses

#730, #450


Stefanuk12:
Some test cases do fail, not sure why...

Stefanuk12:
I've implemented my idea. Not able to get the translations of course but I also modified the tests and such. By default, the original behaviour is seen.

Stefanuk12:
This could be expanded on and allowing control of this behaviour per flashcard instead of a global setting. I'm not sure the best way of doing this so I did not include it.

Stefanuk12:
There's currently a bug where the scheduling data is not included within the card text. I need to add a look ahead for that.

PaperOrb:
Can't wait for this to be merged because adding a newline before and after every piece of text that I want converted into a card is ugly and tiresome.

PaperOrb:
Does the force-push mean that this was added to the main repository?

Stefanuk12:
> @Stefanuk12  Does the force-push mean that this was added to the main repository?

Nope, I'll close this PR - I forgot to. The new update broke a lot of things so I'll need to work on alternative: implementing anki-type cards

Issue #774, opened 2023-10-24
[BUG] Questions not removed from the card postponement list (aka bury sibling cards)


ronzulu:
**Describe the bug**

(Only with the "bury siblings" option enabled)

1. Questions with only a single card (e.g. a simple "::" but not ":::") when reviewed are added to the postponement list. This typically is not critical, but is confusing if the user manually deletes the schedule info from the note, but doesn't get to review the card again until the next day) - since version 1.10.3
2. Questions aren't being removed from the postponement list - since version 1.10.2

**Expected behavior**

1. Questions with only a single card are not added to the postponement list.
2. Questions should be removed from the postponement list the following day.

**Versions**

-   Plugin version: 1.10.2/1.10.3 (as detailed above)

**Additional context**

A bug I introduced.
Apologies to the team, not my finest work!

 **Status**

I have fixed this and will submit a pull request shortly.

st3v3nmw:
Hi @ronzulu,
I've merged the pull request, I'll do the release over the weekend

st3v3nmw:
I've included it in the release 😊

PR #775, opened 2023-10-24
Bug fix: questions now removed from postponement list; etc

ronzulu:
Fixes: https://github.com/st3v3nmw/obsidian-spaced-repetition/issues/774

1. Questions with only a single card are not added to the postponement list.
2. Questions should be removed from the postponement list the following day.


st3v3nmw (approved):
Thanks for the fix!

PR #733, opened 2023-08-25
Support for many clozures within a single flashcard

Stefanuk12:
# Overview

Currently, a flashcard can only have one cloze and the rest are treated as highlights. This PR hides and treats all possible clozures. The new system is also simpler and allows more characters to be easily added or removed in the future.

Note: I accidentally created a new branch from my previous PR which results in the weird commit history. View the files changed for a better overview...

# Changed behaviour

## Before

*creates two flashcards*

```
many text
such surprise [...] more ==text==
some text after
```

```
many text
such surprise ==wow== more [...]
some text after
```

## After

```
many text
such surprise [...] more [...]
some text after
```

# Addresses

#486

PR #769, opened 2023-10-21
Support for Anki Cloze Syntax

Stefanuk12:
#93

I haven't made the test cases yet and I'm not able to at the moment, but I might be able in the future - or you could?

PR #796, opened 2023-11-21
Support for many clozures within a single flashcard

Stefanuk12:
Duplicate of #733

### DIFF
--- a/src/QuestionType.ts
+++ b/src/QuestionType.ts
@@ -107,29 +107,52 @@ class QuestionType_Cloze implements IQuestionTypeHandler {
         let front: string, back: string;
         const result: CardFrontBack[] = [];
         for (const m of siblings) {
-            const deletionStart: number = m.index,
-                deletionEnd: number = deletionStart + m[0].length;
-            front =
-                questionText.substring(0, deletionStart) +
-                QuestionType_ClozeUtil.renderClozeFront() +
-                questionText.substring(deletionEnd);
-            front = front
-                .replace(/==/gm, "")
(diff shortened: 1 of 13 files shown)


ronzulu:
Hi @Stefanuk12 

I'm wondering if it would be more accurate to call this:
- "Support for all clozures within a single flashcard" ("all" instead of "many")? Or 
- "Hide all clozures within a single flashcard"?

When turning on the new option, **all** clozes are included in the one flashcard. Is that correct?

Also, do you think that a global option is the best way to go? 

I've just noticed your other PR https://github.com/st3v3nmw/obsidian-spaced-repetition/pull/769, which looks really interesting and very flexible. I'll try and have a look in the next week or so.

Is there anything this PR can do that https://github.com/st3v3nmw/obsidian-spaced-repetition/pull/769 cannot?



Cheers
Ronny

Stefanuk12:
Yes, this would effectively remove siblings and place them all in one card.

Anki syntax support would be more flexible and possibly a better solution. The only drawback is the ugly syntax. Using a highlight is a lot more concise and convenient. Others have suggested labelling the highlight with [^1] to enable this PRs functionality so that could also be done. To slightly remedy the ugly syntax, my feature request about the custom reading view could help.

Mainly, it's a tradeoff from flexibility and convenience.

An alternative solution to this PR could be to assign different "methods" to different functionalities. For example, using ** would show the default behaviour and using == would show this PRs behaviour.

st3v3nmw:
We now have support for many clozes through @MotaOcimar's contributions. I'll close this now.

Issue #840, opened 2024-01-13
[FEAT] Command to open the "Review Queue" on the right view panel.


Steindvart:
### Problem
The right view panel "Notes Review Queue" opens only when the plugin is loaded. And if you close it, then, therefore, it will be possible to open it only when the plugin is restarted. There is no command or button to open it in an intuitive way.

![image](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/53004350/01ef1df0-4b21-4706-b9ae-1dd53cfca7ef)

### Solution
Need to add a command to open this panel.

### Context

Actually, I want to realize this by making a contribution. This Issue is more for discussion, if any, and better documentation.

st3v3nmw:
This was released in v1.12.0, closing...

PR #846, opened 2024-01-16
[FEAT] Command to open notes review panel - #840.

Steindvart:
## What does this PR?
1. Implemented #840.
2. Also fixed potential undefined behavior with call `async` functions without `await` in activation of panel.

## Examples
1. Added a new command in commands panel.
  image
2. After choose command Notes Review Queue is correctly opened.
  image

## Backward Compatibility
Is safe. There don't seem to be any changes that could potentially lead to problems here :)


st3v3nmw (approved):
LGTM, thanks!

st3v3nmw:
Looks like there's a missing entry for the new Italian translation that went in recently, I'll merge this and fix that afterwards

PR #869, opened 2024-02-05
Implementation of some little UI improvements

KyleKlus:
This PR implements a few UI improvements, which I discovered while I was using spaced repetition.

- The flashcard menu items where space by a margin and not by using a gap in the flexbox of the parent and thus the items where  slightly shifted to the left by the last margin.

- There was one auto too much on the flashcard response container, so that all buttons within it where slightly shifted to the left again.

-  Then there was a hack with margins on the response container of the cram card mode, which I fixed, so that there is no need for it and also so that the "hard" and "easy" button are centered again.


wrgrant:
Nice!

st3v3nmw (approved):
Thanks!

Issue #872, opened 2024-02-08
[FEAT] small ui overhaul suggestions w/ concept drawings


KyleKlus:
Hi there,

I have some ideas for a small ui overhaul on "obsidian spaced repetition" and I have made some concept drawings. So, I wanted to share these here and see if anybody has some remarks (especially @st3v3nmw, @ronzulu). I would be happy to implement those ideas myself in a PR draft.

### Proposal for code changes

- [x] Refactoring FlashcardModal into several views.
- [x] Refactoring keyboard listener
- [x] Implementing ui changes

### Concept drawings
_left -> mockup, right -> comments_

![image](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/97227530/90c9f4e2-3f64-4514-8d6a-b6ca1f27c6b0)

![image](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/97227530/5da7c10b-af98-4387-9263-543073d6e983)

![image](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/97227530/6db78c69-1ec8-4fa8-859c-33b7dec4850c)

![image](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/97227530/cf2ad959-95e0-4dd9-830f-8eed549c0723)

ronzulu:
Hi again

I like nearly all of your ideas, but unsure of a couple.

(2) Are you planning to include the full word or just the first letter? I'm just concerned about non-English translations where potentially the first letter isn't unique for new/due/total

(3) nice suggestion, but the disadvantage is that it would cause (some, although temporary) confusion for existing users. Also would be confusing if a user was to lookup up existing documentation or posts, and the order was opposite to the software.

I'm sure you are much more visually aware than most :smile: It never crossed my mind that the colours on this page are in a different order to the ones on the buttons on the card review view!

![image](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/75528127/d2929191-a6c3-44f0-ae55-37e934f493ce)

KyleKlus:
(2) I came across this issue too, so I included the whole word, instead of just the first letter. I think it looks worse than just having one letter or nothing in front of it, but I was always confused, which color means what, so I kept it, as it improves the onboarding experience for new user a lot.
This is how it now looks:
[...]

Issue #851, opened 2024-01-21
[BUG] Flashcard review order "random card from random deck" - make info in modal title useful


ronzulu:
**Describe the bug**
From comment by @4Source https://github.com/st3v3nmw/obsidian-spaced-repetition/issues/814#issuecomment-1902236712.

The title information for the flashcard review modal has the format:
`${deckName}`: `${cardCount}`

With:
`${deckName}` - The name of the deck of the currently shown card
`${cardCount}` - The count of cards remaining to be reviewed in this deck.

This is useful info for all review order modes, except for "random card from random deck".

**Expected behavior**
In this mode it is more useful to display:

`deckName` - the name of the deck chosen by the user (doesn't change as cards from different subdecks are shown)
`descendantDeckCount` - count of descendant decks under the deck chosen by the user. This being the remaining count, i.e. the count of descendant decks that have 1 or more remaining cards for review
`totalCardCount` - total number of cards remaining for review across the chosen deck and all descendant decks (I.e. not the count of cards remaining in the current randomly chosen deck).

4Source:
What do you think about a Format 
Title: ``${deckName}: ${cardCount}``
Subtitle: ``Subdecks: ${subdeckCount} | ${subdeckName}: ${subdeckCardCount}``

``deckName`` - the name of the deck chosen by the user (doesn't change as cards from different subdecks are shown)
``cardCount`` - the total number of all cards in subdecks chosen by the user 
``subdeckCount`` - the number of all subdecks with remaining cards to review
``subdeckName`` - the name of the subdeck from which the current reviewed card is
``subdeckCardCount`` - the number of remaining cards of the subdeck from which the current reviewed card is

4Source:
Something like that
![image](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/38220764/30d2d10b-00f6-4580-a136-e741907e723e)

When the lowest level of a deck is the selected one ``subdeckName`` and ``deckName`` would be the same that maybe a downside

KyleKlus:
This should be fixed now, so I'll close the issue.
Please reopen it if this bug still persists😇

Issue #748, opened 2023-09-27
[FEAT]


lacanzizek:
**Is your feature request related to a problem? Please describe.**
sm-2 is outdated, can you please replace it with the fsrs algorithm?
https://github.com/open-spaced-repetition/fsrs4anki

SubZeroX:
Yes, this will be really great!

reinvanimschoot:
Is there any update on this? Is this something that is in the pipeline?

ronzulu:
Hi @reinvanimschoot  thanks for the prompt, I just looked at https://github.com/open-spaced-repetition/fsrs4anki now.

Seems very well researched.

I'll have a look at how easy/difficult it would be to integrate, and post back here hopefully in a week or so.

Cheers
Ronny

reinvanimschoot:
@ronzulu awesome!

MotaOcimar:
Hey guys!  This issue has already been addressed in https://github.com/st3v3nmw/obsidian-spaced-repetition/issues/516.

Additionally, a working implementation was made: https://github.com/open-spaced-repetition/obsidian-spaced-repetition-recall

It seems like an attractive idea to combine the two projects. But I don't know how hard it can be

ronzulu:
Hi @MotaOcimar thanks for letting us know, I hadn't come across it before.

The project you reference https://github.com/open-spaced-repetition/obsidian-spaced-repetition-recall is a fairly up-to-date fork from our project itself.

> It seems like an attractive idea to combine the two projects. But I don't know how hard it can be

Definitely worth investigating. If it doesn't work out, I'm sure looking at `obsidian-spaced-repetition-recall` will make it significantly easier for us to bring FSRS functionality into this project.

Thanks once again!

Newdea:
hi @ronzulu , glad to see you're planning to do this, this branch https://github.com/open-spaced-repetition/obsidian-spaced-repetition-recall works fine for now.

Please let us know if there is anything we can do to help.

If you're going to use this algorithm, the data structure of the cards will need a major refactoring. because I'm modifying it based on the master branch, it's a bit messy in that regard.

ronzulu:
Hi @Newdea thank you so much for your kind offer, and all the work in creating the https://github.com/open-spaced-repetition/obsidian-spaced-repetition-recall branch.

I'm wondering if integrating the FSRS algorithm actually required the alternate `tracked_files.json` data store, or if you just preferred it. 

[...]

PR #873, opened 2024-02-09
Gui refactored into separate views | rewrite of keydown handler into event handler

KyleKlus:
### Description
This PR implements a refactor of the flashcard-modal.tsx into serveral view files (DecksListView, FlashcardReviewView).
This was done to ensure code maintainability for possible later overhauls (eg. issue #872). Like this every view has its separate file, where only the code for this view is put.

Another thing, which this PR implements is the rewritten keydown handler. Now it is written as proper event listener, with extensibility (Through switch statements) and readability in mind.

### Changes

-   Improved how spacing between the buttons was done
-   Centered the button groups, which wheren't centered
-   Fixed a hack in the response of the cram card mode
-   Refactored each view into it's own file, to reduce clutter and having a monolithic file
-   Fixed keydown handler with a proper event handler and a switch


ronzulu:
Hi @KyleKlus thanks so much for doing this, especially with the detailed explanations on #870.

Hopefully I'll get a chance to review over the next couple of days. I kind of got excited about #748, and how much research has gone into the https://github.com/open-spaced-repetition/fsrs4anki algorithm.

Actually, would love your thoughts regarding #851 whilst you are "in the neighborhood"

Cheers
Ronny

KyleKlus:
The new algorithm in #748 really looks nice and I think it could be a great benefit for the users. Maybe it could be first implemented as an opt-in feature so it can be tested first, how it behaves and feels.

I also had the same idea as mentioned in #851; I think it is particularly useful, so that one roughly knows how long the learning session might still take. And it would be very easy to implement it, as all data is already there and the ui has still some space for a subheading. 
The only thing to consider is a possible jump in the ui, if the root deck has cards and then after those are done it switches to the subdeck. The subheading could push the question section down if it appears, so I would pre allocate the space for the sub deck info, so it doesn't push anything down when it appears.

I would be happy to implement that while I am working on the other ui changes.

ronzulu:
> The new algorithm in #748 really looks nice and I think it could be a great benefit for the users. Maybe it could be first implemented as an opt-in feature so it can be tested first, how it behaves and feels.

I was thinking the same 👍 

> I also had the same idea as mentioned in #851, but just for the normal review mode; I think it is particularly useful, so that one roughly knows how long the learning session might still take.

Good point, I hadn't thought of that

> The only thing to consider with the subheading is a possible jump in the ui, if the root deck has cards and then after those are done it switches to the subdeck. The subheading could push the question section down if it appears, so I would pre allocate the space for the sub deck info, so it doesn't push anything down when it appears.

I think I understand your point, personally it wouldn't bother me if occasionally after completing a card during the review, the next card is displayed at a slightly different vertical position because of the extra info. I only use Obsidian on the desktop so vertical space isn't a premium, as long as on a mobile the pre allocated space doesn't seem wasted when it's blank.

> I would be happy to implement that after I've made the other ui changes in #872 😊.

Awesome!

KyleKlus:
Hi again, I am done now with all the ui improvements I wanted to do in #872. The PR for the this would build upon this unmerged PR, as I finalized/improved the view switching a bit there 🚀 [Link to the changes](https://github.com/KyleKlus/obsidian-spaced-repetition/tree/ui-overhaul).

So would the reviewing process be easier if I wait until this one is finished or would it be better, if I close this one and open another one with all the ui changes at once?

I just don't want to overwhelm the review process with too many changes 😇.

ronzulu:
Hi @KyleKlus, I like your idea to make the UI changes under a separate PR. That way this PR is purely for the refactoring. Cheers

docs/changelog.md
```diff
@@ -17,6 +17,8 @@ Generated by [`auto-changelog`](https://github.com/CookPete/auto-changelog).
 -   Improved how spacing between the buttons was done
```
ronzulu, line 17: Nice detailed change log.
On this project the convention seems to be simply have the name, and issue or PR link.
Perhaps move the details to the PR description.

src/gui/DecksListView.tsx
```diff
@@ -0,0 +1,170 @@
...
+import {
+    DeckStats,
+    IFlashcardReviewSequencer as IFlashcardReviewSequencer,
+} from "src/FlashcardReviewSequencer";
+import { TopicPath } from "src/TopicPath";
+import { FlashcardModalMode } from "./flashcard-modal";
+
+export class DecksListView {
+    public plugin: SRPlugin;
+    public mode: FlashcardModalMode;
```
ronzulu, line 19: I don't think this is needed as it is always set to FlashcardModalMode.DecksList

docs/changelog.md
KyleKlus: I moved this now into the PR.
(Sry, that it took me a while, but I had to write some uni exams this week)

src/gui/DecksListView.tsx
KyleKlus: I removed this now from the class.

KyleKlus:
I think, I fixed now all the warnings. Sry, that I didn't removed the unused variables; I don't know why I forgot that 😅.

st3v3nmw (approved):

st3v3nmw:
LGTM, thanks!

PR #874, opened 2024-02-11
fix card isNew

Newdea:
when card has dummy schedule, this isNew return false, but should be true. current is 
```ts
get isNew(): boolean {
        return !this.hasSchedule;
    }
```
change to
```ts
get isNew(): boolean {
        return !this.hasSchedule || this.scheduleInfo.isDummyScheduleForNewCard();
    }
```

### DIFF
--- a/src/Card.ts
+++ b/src/Card.ts
@@ -21,11 +21,11 @@ export class Card {
     }
 
     get cardListType(): CardListType {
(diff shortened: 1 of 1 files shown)


ronzulu (approved):
Thanks for finding this problem and fixing it. LGTM.

Issue #747, opened 2023-09-23
[FEAT] Code integration


DanielChicoUGR:
Modify the parser to search inside the code blocks. With the use of the [Admonition](https://github.com/javalent/admonitions) plugin, the text inside is not detected as flashcard.

ronzulu:
Hi there, can you include some sample Markdown.

DanielChicoUGR:
Here is what i mean.
![Screenshot_2023-09-25-11-47-35-723_md.obsidian.jpg](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/81250900/362a75d2-63e1-4cfd-bc84-3c42324aa4cf)

![Screenshot_2023-09-25-11-47-45-987_md.obsidian.jpg](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/81250900/e259f506-0380-4f17-a586-e8238ff2b7e0)


Thats part of my Fundaments of networks homework

I have looked the code of the parser but i dont know where to edit, (obviusly in the parser file....)

ronzulu:
Hi there

The parse function seems to explicitly skip over the content of code blocks, anything starting with ` ``` ` or `~~~`.

I can imagine that in general that makes sense, code could easily have `::` within a line for example,

A couple of options come to mind for implementation

1. Include a new user option which would specify which types of code blocks should be parsed for questions.
2. Some kind of markup within individual code blocks... perhaps just some variation of the standard  
`<!--SR:!2023-09-20,1,230--> `

Cheers
Ronny

DanielChicoUGR:
OOOO u are right, ill try something and i will make a pull req if i have something interesting

reinvanimschoot:
Hey guys, is there any update on this? Or any workaround? I made an admonition called `sr-cards` only to realise that the cards are not picked up inside it haha. But it is much cleaner to have a folded admonition to store the cards, so I'm hoping for a workaround!

DanielChicoUGR:
Next week i end the smester and try some thoughts 😅😅😅😅

DanielChicoUGR:
Sry for the delay, what i need to set up the deverlopment enviroment???
If never contributed to an open source project and i would be nice to have a littel guide to follow and dont mess up the repo settings.

PD i know how to yuse github at a medium level, i just need to know if i need node or something like that and the protocol to make a pull request with all the info.

DanielChicoUGR:
[...]

PR #885, opened 2024-02-24
Added support for flashcards within Admonition style code blocks

DanielChicoUGR:
## Description:
As described in issue #747 , support for flashcards within Admonition style code blocks has been added. The plugin is now able to detect cards inside code blocks with ad-* style indicator, among impklemented in [ad-monition](https://github.com/javalent/admonitions) pluggin

image

image

image

image

image

## Changes:
- Modification of the parser function to skip those blocks of code that do not begin with: "```ad-" or "~~~ad-"


ronzulu:
Hi @DanielChicoUGR, nice one!

I haven't had a chance to test this, but a couple of comments in the meantime:

Please include new unit test cases within `parser.test.ts`

About terminology, IMHO saying "Added integration with the Admonition Pluggin" implies actual integration with that plug-in, which isn't accurate. I think "Added support for flashcards within Admonition style code blocks" is more accurate.

There is a chance that the enhancement could break existing notes. This would occur if someone had included within an admonition code block something that matched any of the recognized card formats. Something that they didn't want to be treated as a card, and up till now hasn't been, e.g. `typedef std::vector<Board> t_bvector; `

Perhaps we need to make this new behavior optional.

Cheers
Ronny

DanielChicoUGR:
kk, U right in both cases, i will work on this today and add some unit test too.

One Question only, is it possible to modify this pr or may i reopen other pr and close this?? (Noob doubts sry)

 thanks for the time and attention, cheers Daniel

ronzulu:
Hi Daniel, looking forward to the update :smile:

Do you have an edit button at the top of this page? I see the following:

image

Recently I was granted "collaborator" permissions, that's why I can edit someone else's PR.

I assume that before I became a collaborator that I would have been able to edit any PR that I opened, but don't explicitly remember.

If you can't edit, then I think it's better if you let me know what changes you want and I will make them (easier than closing this PR and opening another)

Cheers
Ronny

KyleKlus:
Hi, I am closing this, as there has been no development on this and the code has already changed quite a lot, so it would have to be reimplemented 😇.

PR #899, opened 2024-03-11
[Overhaul] Implementation of a more consistent & modern UI

KyleKlus:
### Description

This PR is an overhaul of the main ui and it implements the issue [`#872`](https://github.com/st3v3nmw/obsidian-spaced-repetition/issues/872). More detail on what the PR implements and more documentation in general can be found there.

### Changes

-   Improved DeckListView
-   Improved FlashcardReviewView
-   Improved EditModal
-   Added a bunch of classes for styling
-   Renamed gui files such that they follow a uniform naming convention
-   Improved mobile  view
-   Fixes key listener bug from issue #907 
-   Fixes bug from issue #773

PR #912, opened 2024-03-20
style(card review modal): reorder stats containers

st3v3nmw:
Small follow-up on #899

I think that we should show the "most useful" counts first. When reviewing, it's best if users see the due & new cards first as this makes it a bit easier to select the deck to review especially for left-to-right languages which are the vast majority of what we support. Or if they have many decks.

@KyleKlus


KyleKlus:
Yes, that is a very valid argument. I just wanted that the colors are in the order of the hard, good and easy buttons, so it is a bit more uniform.

If I just switched the colors, then the association of the user, which color represents what would maybe confuse some users.

But I am fine with them switching back, if you choose function over form here.

Issue #915, opened 2024-03-22
[BUG] Plugin not picking up certain flashcards


bgreenawald:
**Describe the bug**
The plugin suddenly only picks up ~half of the flashcards in my Vault. As of a few days ago, I had nearly 500 flashcards, but when I loaded the flashcard review today, it only showed that I had around 250 flashcards. I can confirm that all the other flashcards are still in the Vault and found an example of a file with flashcards past their review date that weren't being picked up at all. I currently don't see any pattern between what it picks up and what it doesn't.

**Versions (please complete the following information):**

-   OS: Windows 11, Android
-   Obsidian version: 1.5.11
-   Plugin version: 1.12.1
-   If on desktop, Installer version: 1.4.13

ronzulu:
Hi @bgreenawald 

Thanks for letting us know.

Could you please post (as attachment, not copy/paste) a markdown file that contains flashcard(s) that don't appear in the review modal.

Also, please attach the plugin's settings file, located e.g.
`D:\Obsidian\.obsidian\plugins\obsidian-spaced-repetition\data.json`

Cheers
Ronny

bgreenawald:
Hi Ronny,

Thanks for the quick response. Attached is an example file that isn't showing up. Here is a picture of the current review screen:
![image](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/8428972/7eb5a242-e9ae-4034-b206-994d0237e29f)

Best,
Ben

[Avicenna.md](https://github.com/st3v3nmw/obsidian-spaced-repetition/files/14729355/Avicenna.md)
[data.json](https://github.com/st3v3nmw/obsidian-spaced-repetition/files/14729370/data.json)

Garg0yle:
I'm having the same issue all of a sudden. From about 50-60 flashcards in my vault to 2. Here are attachments. Thanks for your work on this plugin btw

[The Taliban at war_ 2001-2018.md](https://github.com/st3v3nmw/obsidian-spaced-repetition/files/14733369/The.Taliban.at.war_.2001-2018.md) 

[data.json](https://github.com/st3v3nmw/obsidian-spaced-repetition/files/14733377/data.json)

Yury-Dictum:
Hi! Same problem here. Two days ago there used to be around 130 cards in the queue, now there are none.

[...]

PR #920, opened 2024-03-27
fix bug with recognizing frontmatter topic tags (led to missing cards shown for review)

ronzulu:
This addresses [ [BUG] Plugin not picking up certain flashcards #915](https://github.com/st3v3nmw/obsidian-spaced-repetition/issues/915)


st3v3nmw (approved):
Thanks for the fix!

Issue #878, opened 2024-02-15
[FEAT] Refactor code to support diff methods of storing the scheduling info, and diff SR algorithms


ronzulu:
The current code has a structure tied to the method of storing scheduling info within the note markdown file itself.

Refactoring the code is a prerequisite to implementing other methods, e.g. as requested in:
[[FEAT] Stop using YAML; Move plugin info and data to separate file](https://github.com/st3v3nmw/obsidian-spaced-repetition/issues/162)

pikatwinky:
It's only my opinion but,
Talking about the note review mode... i think having the note schedulling info in the frontmatter is useful when using combined with Dataview, so you can create through the use of queries custom review sessions to filter down only certain set of notes based on properties (in particular using the 'due date' field), functionality none existent currently in the plugin (you have to review your notes in a global way) 
But on the other hand, talking about the flashcards, i think having their scheduling info in a separate db could help to unclutter the notes, but of course, maybe this last is unfeasible

YIRU69:
I hope you can bring FSRS functionality into this project!

ronzulu:
I hope so! I've completed implementing the re-factoring addressed by this issue, but not yet completed testing.

After this then there are a couple of other nontrivial tasks to complete...

YIRU69:
Hi! Can you tell me approximately how long until I can use FRSR?

ronzulu:
Hi @YIRU69 if I'm the one implementing, I'd think realistically say 6 months, so around October this year.

st3v3nmw:
@ronzulu this is done now right?

ronzulu:
Hi @st3v3nmw 

More will need to be done when new data stores are actually implemented, for example converting from one Datastore type to another.

But I'm happy enough with what's been completed so far, so I'll close this now.

YIRU69:
@ronzulu Hi, glad to hear you complete this huge work. But it just part work for FSRS? I check the newest update and can't find FSRS.

ronzulu:
Unfortunately @YIRU69 that is correct. It's only one part.

Also, I've taken @danhje suggestion and am first working on [ [FEAT] user defined "Easy, Good, Hard" values instead of or in addition to the algorithm defined one. #741 ](https://github.com/st3v3nmw/obsidian-spaced-repetition/issues/741).

The rationale being that #741 is a practical and easier way of verifying the work done here in #878 compared with FSRS. (Apart from all the unit test cases I added here...)

PR #952, opened 2024-05-08
Add negative tags

bitesizing:
**This PR adds support for 'negative' tags when choosing which whole notes should be reviewed.**

This was added because it is sometimes useful to exclude notes with a specific tag from review.

For example, I use spaced repetition to review notes associated with academic papers, tagged with `#paper`. However, I sometimes create these notes before reading the paper so that I can link them to other notes; in these cases they are tagged `#unread` and I do _not_ want them to show up for review, but I also want them to keep the paper tag.

image

Now you can ignore a subset of notes by adding a 'negative' tag to the settings, formatted like `-#unread` (shown above). This gives users more options if they want to review notes _using an existing tag system_, instead of having to manually cultivate a tag just for reviewed notes. 

[[Edit: I rewrote this to better document the changes and explain why they're useful]]


ronzulu (changes requested):
Hi @bitesizing, thanks for this.

Since you implemented this, the code base (beta) has changed considerably, primarily due to:
[ Feat 878 Refactor code to support diff methods of storing the scheduling info, and diff SR algorithms #1006 ](https://github.com/st3v3nmw/obsidian-spaced-repetition/pull/1006)

1. So I think we need to extract the logic that you implemented and move to a new spot, possibly something in the new `SettingsUtil` class. There is an existing method `filterForNoteReviewTag` that may be of interest.

The latest beta code is available at:
[Beta release of outstanding PRs #1011](https://github.com/st3v3nmw/obsidian-spaced-repetition/discussions/1011)

2. We also should add a couple of unit test cases.

3. I wonder if there are better terms positive/negative. For example perhaps `exclusion tag` conveys the meaning better than negative.

bitesizing:
Hi @ronzulu, thanks for this! It seems like you've done loads of impressive work on the repo:)

Just to be clear because I'm quite new to contributing, are the changes to the code base that you mentioned reflected in the newest pushes to the master branch? I.e., can I just fork the repo and reintegrate my changes into the new structure. Or do I need to access the source code for the beta release directly?

ronzulu:
Hi @bitesizing , thanks for the feedback :-)

This is new territory for me as well actually. The changes to the code base haven't been integrated into the master repository `https://github.com/st3v3nmw/obsidian-spaced-repetition`. Unfortunately the project owner @st3v3nmw has been out of contact, so we don't know when this would happen.

I've merged in a few different outstanding PRs into the "beta code" on my repository. I was able to do that without involving the PR contributor as those changes affected different parts of the code.

So this is the first time your question has come up. I guess there are 2 options:
1. Wait and see when Stephen is back on board and has integrated other PRs including #878 into `st3v3nmw/obsidian-spaced-repetition`. and then git update your branch from it.
2. Fork from my repository, and make the necessary changes to incorporate your work. I would then merge your changes into my "beta branch".

If you choose 2, let me know if you have trouble as no one has needed to do this before.

Cheers
Ronny

Issue #934, opened 2024-04-17
[BUG] Arrows in note review panel don't move (in default blue toppaz theme)


arias007:
describe the bug
When using the default blue toppaz theme, the up and down slide arrows don't slide with the file information.

To Reproduce
Go to 'spaced repetition on right'
Click on '....'
Scroll down to 'botton' and unfold
See error：arrow freezed on screen while scrolling
Expected behavior：arrow scroll with filename

Screenshots
If applicable, add screenshots to help explain the bug. This could be the logs in the developer console (CTRL + SHIFT + i).
![image](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/90596315/32e6dba2-efa5-4074-84a0-be5dd9296be6)

Versions (please complete the following information):

OS: [e.g. Android&windows]
Obsidian version: [e.g. v1.15.2]
Plugin version: [e.g. v1.12.3]
If on desktop, Installer version: [run this.navigator.appVersion on the developer console, e.g. 5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) obsidian/0.12.3 Chrome/89.0.4389.128 Electron/12.0.6 Safari/537.36"]

ronzulu:
Hi @arias007, please edit the issue description to complete the required information:

![image](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/75528127/2294ab7e-9a27-4226-b053-8bd369a75838)

arias007:
ok，thanks

arias007:
describe the bug
When using the default blue toppaz theme, the up and down slide arrows don't slide with the file information.

To Reproduce
Steps to reproduce the behavior:

Go to 'spaced repetition on right'
Click on '....'
Scroll down to 'botton' and unfold
See error：arrow freezed on screen while scrolling
Expected behavior
arrow scroll with filename

Screenshots
If applicable, add screenshots to help explain the bug. This could be the logs in the developer console (CTRL + SHIFT + i).
![image](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/90596315/32e6dba2-efa5-4074-84a0-be5dd9296be6)


Versions (please complete the following information):

OS: [e.g. Android&windows]
Obsidian version: [e.g. v1.15.2]
Plugin version: [e.g. v1.12.3]
[...]

PR #962, opened 2024-05-22
fix: arrows in note review panel don't move

4Source:
I had a look at this bug #934 and implemented a fix. Just a few changed where required. 
Mostly I changed the classes the nav items should have to match the structure of obsidian file hierarchy. 

- [x] Test run successfully 
Test Suites: 19 passed, 19 total
Tests:       206 passed, 206 total
- [x] Fixes Issue


src/gui/Sidebar.tsx
```diff
@@ -173,18 +171,8 @@ export class ReviewQueueListView extends ItemView {
         }
 
         folderTitleEl.onClickEvent(() => {
-            for (const child of childrenEl.childNodes as NodeListOf<HTMLElement>) {
-                if (child.style.display === "block" || child.style.display === "") {
-                    child.style.display = "none";
-                    (collapseIconEl.childNodes[0] as HTMLElement).style.transform =
-                        "rotate(-90deg)";
-                    deck.activeFolders.delete(folderTitle);
```
4Source, line 181: What is this for?

4Source: this line: deck.activeFolders.delete(folderTitle);

ronzulu: This stores the collapsed/expanded state of each folder (folder names being things like
    // "TODAY", "NEW" or formatted dates).

So here where the tree node is being collapsed, the folder name is being removed

4Source: Is the call for this function at this spot correct it seems for me not necessary?

ronzulu: Like me, do you also think that the `activeFolders` is used to maintain the state of which folders are expanded?
This code is being executed when the user clicks on an expanded folder name, causing it to collapse.
So it seems to me that calling `deck.activeFolders.delete(folderTitle);` is valid.

4Source: Yeah I think your right I will fix this

st3v3nmw (approved):
LGTM

PR #972, opened 2024-05-24
[FIX] Folder ignore sorts all folder starting with string

### LINKED ISSUE #969: [BUG] Folder ignore sorts all folder starting with string
**Describe the bug**
I recently discovered that if you have brunche of  folders there are starting with same text e.g. ``Test-1`` ``Test-2`` and you put ``Test-`` to ignore list both get ignored.

**To Reproduce**

1. Create 2 folder ``./Test-1`` ``./Test-2``
2. Create in each a file with test flashcard
3. Add to folder ignore list ``Test-``

**Expected behavior**
Only exact match of the name should be ignored

-   OS: Win10
-   Obsidian version: 1.5.12
-   Plugin version: 1.12.4
-   If on desktop, Installer version: 5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) obsidian/1.4.16 Chrome/114.0.5735.289 Electron/25.8.1 Safari/537.36

**Additional context**
[TestFolders.zip](https://github.com/st3v3nmw/obsidian-spaced-repetition/files/15421853/Test.zip)

4Source:
Fixes #969 

Implementation of better comparison 
Implementation of Unit Test

### DIFF
--- a/src/util/utils.ts
+++ b/src/util/utils.ts
@@ -110,10 +110,10 @@ export function stringTrimStart(str: string): [string, string] {
  * if toCheck = "root/s" -> false
  */
 export function isEqualOrSubPath(toCheck: string, rootPath: string): boolean {
-    const rootPathSections = normalize(rootPath.toLowerCase())
+    const rootPathSections = normalize(rootPath.toLowerCase().replaceAll(/(\\\\|\/)/g, sep))
         .split(sep)
         .filter((p) => p !== "");
-    const pathSections = normalize(toCheck.toLowerCase())
+    const pathSections = normalize(toCheck.toLowerCase().replaceAll(/(\\\\|\/)/g, sep))
         .split(sep)
         .filter((p) => p !== "");
     if (pathSections.length < rootPathSections.length) {
--- a/src/util/utils.ts
+++ b/src/util/utils.ts
@@ -110,10 +110,12 @@ export function stringTrimStart(str: string): [string, string] {
  * if toCheck = "root/s" -> false
  */
 export function isEqualOrSubPath(toCheck: string, rootPath: string): boolean {
-    const rootPathSections = normalize(rootPath.toLowerCase().replaceAll(/(\\\\|\/)/g, sep))
+    const rootPathSections = normalize(rootPath.toLowerCase())
+        .replaceAll(/(\\|\/)/g, sep)
         .split(sep)
         .filter((p) => p !== "");
-    const pathSections = normalize(toCheck.toLowerCase().replaceAll(/(\\\\|\/)/g, sep))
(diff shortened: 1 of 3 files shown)


4Source:
Not sure why the Test Failed for me tried it multiple times no issues with this. Maybe because Test is running on Ubuntu and i am on windows I will check I can Test this on a Virtual machine.

4Source:
Finally Test succeed

ronzulu (approved):
LGTM @4Source 

Not that it matters too much, but I think we use different terminology in terms of what is "sub". I guess I think of "sub" to encompass less than the original thing, e.g. "subsection".

If I've understood correctly, then I think you have the opposite meaning. That "root" is a subpath of  "root/sub/sub2", but encompasses more.

As an idea, what do you think of "isEqualOrAncestor" instead?

Great that you've included comprehensive test cases.

4Source:
Actually I think we have the same understanding of sub.
> Not that it matters too much, but I think we use different terminology in terms of what is "sub". I guess I think of "sub" to encompass less than the original thing, e.g. "subsection".

Where do you mean is this the case?
> If I've understood correctly, then I think you have the opposite meaning. That "root" is a subpath of "root/sub/sub2", but encompasses more.

I am not a native speaker so If you say this fits more than I can change it
> As an idea, what do you think of "isEqualOrAncestor" instead?

src/util/utils.ts
```diff
@@ -96,6 +97,38 @@ export function stringTrimStart(str: string): [string, string] {
...
 
+/**
+ * Checks a path is equal or a subpath of the other rootPath
+ *
+ * @param toCheck The path to check it is equal or a subpath of path.
+ * @param rootPath The ref path to check the other is equal to or a subpath of this.
+ * @tutorial
+ * rootPath = "root/sub/sub2"
+ * if toCheck = "notRoot/..." -> false
+ * if toCheck = "root" -> true
```
ronzulu, line 108: I think I understand the confusion now. Yes we do have the same understanding of the word "sub", this might just be a documentation mistake?

`if toCheck = "root"` -> `true`
`if toCheck = "root/sub"` -> `true`

Should those be false? I think one of the test cases checks that it is false
`expect(isEqualOrSubPath(root, rootPath)).toBe(false);`

ronzulu: BTW your English is great!

4Source: Yes I think your right the documentation seems to be wrong. And now I understand what you mean with root path is not the correct naming. With ``rootPath``  I meant the path and all of its sub paths but It is not actually the root path. 

But the test should be correct if the path to check is above the ``rootPath`` it is not a part of the sub folders.

PR #751, opened 2023-09-30
refactor: Separate business logic from user interface code

ronzulu:
Based on https://github.com/st3v3nmw/obsidian-spaced-repetition/releases/tag/1.10.1

- Separated out "business logic" from the user interface code for flashcards, as well as other refactoring.
- The intention has been to keep existing functionality without change, but structure the code to support future contributions.
- As part of the restructure, many unit test cases were added - now around 170 (up from 20)

### Bug Fixes

To simplify the process of reviewing for the project owner, apart from keeping functionality the same, no bugs have been fixed. However, during the restructure and testing, it was noticed that the following was fixed:

-   [Fixed] Revisons cannot be saved properly [`#745`](https://github.com/st3v3nmw/obsidian-spaced-repetition/issues/745)
-   [Fixed] Decrease in Card Count after Pressing 'Reset'  [`#704`](https://github.com/st3v3nmw/obsidian-spaced-repetition/issues/704)

### Details

-   Development: Thoughts about implementation redesign? [`#736`](https://github.com/st3v3nmw/obsidian-spaced-repetition/discussions/736)

### Commit Comments

Much work went into this restructure. Until the home stretch, the commit comments weren't used (just "something")
There wouldn't be much to gain by examining any of the interim commits.

Issue #741, opened 2023-09-16
[FEAT] user defined "Easy, Good, Hard" values instead of or in addition to the algorithm defined one.


visaeryon:
**Is your feature request related to a problem? Please describe.**
Yes it is, I use real short 'maximum interval' like 7days or 8days this makes the recommended "easy, good, hard" intervals to saturate real quick by that I mean, all the three of them will show the max interval after some time usually after 3-4 consecutive "easy or good" reviews for a given card, I don't really want that to happen and I think this defeats the purpose of reviewing a given card altogether.

**Describe the solution you'd like**
I want the increment intervals ie., "easy, food, hard" to be constant values and is user defined in the settings, like 7days for easy, 4days for good and 1day for hard, and this should be constant for all the cards in a deck no matter new or old ie., bypassing the algorithm altogether. I'm not saying get rid of the algorithm, this is more like an alternative option for people like me who don't get how the algorithm works.

**Additional context**
I really don't get how the algorithm works, the interval increments are almost always partial values like 3.5days, 7.3days, I tried various combinations of 'base ease' and 'interval' values in the algorithm settings but it doesn't seem to do much, on top of it there is this saturation problem with the algorithm. I also don't want the values of the review intervals to be constantly changing from card to card and depending on the previous reviews, I just like constant intervals it makes the reviewing process a bit easy, I don't have to check for the new easy intervals for each and every card I review, I review a lot of them, trust me it becomes tedious after some time, and it also breaks the flow of recalling the content that I'm reviewing, if I knew beforehand that "easy" will mark this card for 6days or 2days for a hard one, then the whole reviewing process will become much more fluid and less tedious.

**here is examples of saturation problem**
![Screenshot (62)](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/70768824/8f2458ed-1284-4bb0-b49c-eb817f153678)
![Screenshot (66)](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/70768824/3927f374-b164-4446-b3fc-fb041a02496d)
![Screenshot (68)](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/70768824/48946344-9e29-4af1-8368-9b571185e0d3)

visaeryon:
[...]

Issue #162, opened 2021-06-16
[FEAT] Stop using YAML Move plugin info and data to separate file


fredriksy:
**Is your feature request related to a problem? Please describe.**
Currently running the plugin and studying is great but it runs havoc on my version control system because of the file changes on everything I studied (due to YAML). This caused due to the storage of spaced repetition data on each individual file via YAML tags.

**Describe the solution you'd like**
A modular approach would be ideal so that the notes are independent of my studying. My notes on "Newton's Laws" shouldn't change every time I just read them right? 
If possible the data on my studying should be stored in a separate container file and that should change. It would make it much easier to track from a version control standpoint how much studying I'm doing.

**Additional context**
To be clear I'm talking about the YAML tags below
![image](https://user-images.githubusercontent.com/8610466/122140281-ddc2cc00-ce18-11eb-94bf-3023947de9e3.png)

st3v3nmw:
Which VCS are you using?
I think conflicts are less problematic with the current system compared to in one central file. For instance, we wouldn't store the file name, but rather some unique id to protect against file renames & stuff. Now trying to make corrections to the central file becomes much more difficult.

fredriksy:
Thanks for the response! 

I use git, I realized my language made things confusing. When I mentioned it caused havoc on my version control system, I meant that it caused change in files that upon review of my git history , hinted that something changed (when in reality, nothing in the content meaningfully changed).

In my head, I'd prefer that all information related to my review would be consolidated in one file. I can easily deal with some messiness with that file. It also seems attractive to be able to look review status in one file.

Your point makes sense, it does seem to be quite difficult to have your plugin somehow track all renames too. Perhaps my recommendation is completely impractical then. 


Feel free to mark this as closed if the task is too hard.

Signynt:
I personally would also find this a valuable feature! For an example on implementation see the [Recall plugin](https://github.com/martin-jw/obsidian-recall).

My reason is that I share a common wiki/notes with friends from med school, that we sync over GitHub, and would like to use Spaced Repetition to review notes. With it in the frontmatter everyone would have the same status of review, so it doesn't work out. This would be solved if the review status was saved in a separate file that would be different for each person.

Newdea:
+1 
want  this a valuable feat too.

TolgaKerem07:
I would also find this useful please add this

L-M-Sherlock:
[...]

PR #1006, opened 2024-06-25
feat: refactor code to support diff methods of storing the scheduling info, and diff SR algorithms

ronzulu:
Hey there, finally have implemented and tested these non-trivial changes - and is now ready for review and beta testing.

https://github.com/st3v3nmw/obsidian-spaced-repetition/issues/878

From the user perspective, there shouldn't be any functional change. But the architecture has changed as follows:

# Code Changes

### Algorithms

The code for implementing the current algorithm has been extracted from the various files and placed in a new class `SrsAlgorithm_Osr` that implements a new interface `ISrsAlgorithm`.

Future algorithms would similarly need to implement the interface `ISrsAlgorithm`.

### Data Stores

The existing code for saving the scheduling info (both for notes in the frontmatter, as well as the HTML comments for the flashcards) have been placed in a new class `DataStore_StoreInNote` that implements a new interface `IDataStore`.

Methods for storing the schedule info in alternate ways would similarly need to implement the `IDataStore` interface.

### DataStoreAlgorithm

Although most of the existing code could be classified either algorithm or data store, some good not. This code was specific to the algorithm and data store combination.

That code was extracted and placed in a new class `DataStoreInNote_AlgorithmOsr` that implements the new interface `IDataStoreAlgorithm`

### Separation Of Core Logic From User Interface Code

https://github.com/st3v3nmw/obsidian-spaced-repetition/pull/751 (about 9 months ago) separated out much business logic from user interface code. Similar change was required in order to achieve the above.

### Unit testing

With more code separated from UI code, it was possible to include more unit test cases. About 30 more test cases were written.

# Next Steps

[ [FEAT] user defined "Easy, Good, Hard" values instead of or in addition to the algorithm defined one. #741 ](https://github.com/st3v3nmw/obsidian-spaced-repetition/issues/741) - Suggestion by @danhje as an easy way to further test these changes

[ [FEAT] Stop using YAML; Move plugin info and data to separate file #162 ](https://github.com/st3v3nmw/obsidian-spaced-repetition/issues/162) - Possibly necessary for supporting algorithms that require the storage of extensive scheduling info

[ [FEAT] sm-2 is outdated, can you please replace it with the fsrs algorithm? #748 ](https://github.com/st3v3nmw/obsidian-spaced-repetition/issues/748)

# Beta testing (help wanted!)

As you can appreciate, lots of work went into this. And even though more test cases were written, it's more than likely I inadvertently broke something (e.g. in functionality that I don't use myself).

[main_878A.zip](https://github.com/user-attachments/files/15966849/main_878A.zip)

@st3v3nmw would be great if you could have a look


ronzulu:
Hi @st3v3nmw , hope you are doing okay.

If you could please review this so that we can merge it into master.

Thanks
Ronny

ronzulu:
Hi @st3v3nmw 

Great that I can merge this into master myself now. But probably better if you can review first.

Cheers
Ronny

image

ronzulu:
Hi @st3v3nmw, just wondering when you will be able to review and merge this. Or are you happy for me to do that?

tests/unit/scheduling.test.ts
```diff
@@ -27,81 +52,115 @@ test("Test reviewing with default settings", () => {
...
         interval: 3,
     });
-    expect(dueDates).toEqual({
+    /* expect(dueDates).toEqual(new DueDateHistogram({
         0: 1,
         1: 1,
         2: 1,
         3: 5,
-    });
+    })); */
```
st3v3nmw, line 122: Has the scheduling behavior changed? I've tried this locally and the commented assertion is now failing. Same for the other assertions that have been commented out in `tests/unit/scheduling.test.ts`.

.github/workflows/release.yml
```diff
@@ -15,6 +15,8 @@ jobs:
 
         steps:
             - uses: actions/checkout@v3
+              with:
+                  node-version: "20"
```
st3v3nmw, line 20: This doesn't seem to be necessary here. It's already covered in the next step (`Use Node.js`)

src/DueDateHistogram.ts
```diff
@@ -0,0 +1,137 @@
...
+        this.dueDatesMap.set(days, value + 1);
+    }
+
+    decrement(days: number): void {
+        let value: number = 0;
+        if (this.dueDatesMap.has(days)) {
+            value = this.dueDatesMap.get(days);
+            this.dueDatesMap.set(days, value - 1);
+        }
+    }
```
st3v3nmw, line 68: More like a question, but why the "asymmetry" between `increment` & `decrement`... `increment` does the `.set` outside the `if` but `decrement` does it inside

st3v3nmw (changes requested):

tests/unit/scheduling.test.ts
ronzulu: Definitely the intention has been to keep scheduling (all functionality really) identical to the current behaviour. I'm not sure what happened here, but I will look into it.

src/DueDateHistogram.ts
ronzulu: It is valid for `increment` to be called when there is no entry in the map related to `days`, in which case we need to set the value to 1 outside the `if`.

But it's not valid for `decrement` to be called in that situation, because we should never be setting the value to -1. I didn't want to throw an exception and just guarded it with the `if`. I'll modify the code to also catch the case of `decrement` being called where `dueDatesMap.get(days)` has a value of 0, and not just if `dueDatesMap.has(days)` is `false`.

tests/unit/scheduling.test.ts
ronzulu: I had modified `osrSchedule` to treat `dueDateHistogram` as an input only (and to not modify it as well). I commented out the test case temporarily, and forgot to fix afterwards. Corrected now.

.github/workflows/release.yml
ronzulu: At one stage during development, something failed (can't remember the details), and I needed to add that to fix the problem. I got rid of those 2 lines now, and recommitted, and all the tests pass. So I'm happy to leave them out.

tests/unit/scheduling.test.ts
st3v3nmw: @ronzulu, I'll try see what has changed as well. For now this is good to go in. There's a merge conflict though

st3v3nmw (approved):

ronzulu:
Hi @st3v3nmw I've updated this branch from master, and rectified the small number of conflicts.

I know that I can click "squash and merge" but thought I would leave it to you :-)

Cheers
Ronny

ronzulu:
Hi @st3v3nmw there are a few PRs ready to merge and release.

I can release them as Beta, but only if they are first merged into master.

So I'll merge this and probably others into master now.

Cheers
Ronny

PR #1012, opened 2024-07-07
Added new option allowing for multiline cards with empty lines

alberti42:
Added `multilineCardEndMarker` setting storing a delimiting string. The user can choose for example '---'. This allows handling flashcards with empty lines, without resorting to expedients such as hidden characters to mimic the empty line.

Bildschirmfoto 2024-07-07 um 15 49 32


ronzulu (changes requested):
Hi @alberti42 thanks for this, many people have been after this for a long time, and will be very happy when it is released.

A few requests:

1. Currently some of the existing unit test cases fail. Please update your enhancement so that the existing tests all pass. See https://www.stephenmwangi.com/obsidian-spaced-repetition/contributing/#code for details
2. Update the user facing documentation, see https://www.stephenmwangi.com/obsidian-spaced-repetition/contributing/#documentation
3. Let me know and I'll review at that time
4. We will need some new unit test cases. You may prefer to wait till I complete step 3. and provide feedback about functionality.

Thanks once again!
Cheers

ronzulu:
Hi @alberti42, I've just noticed that you have rearchitected the parser from the ground up. You must have put in much time and thought into this!

@4Source initiated the discussion [Rework the flashcard seperator/detecting system #1018](https://github.com/st3v3nmw/obsidian-spaced-repetition/discussions/1018) noting that there are many outstanding requests that relate to the parser and that it's not very easy to enhance in its current form. 
(The list including the issue titles at https://github.com/users/ronzulu/projects/1)

So your rearchitect of this part of the plugin might be perfect!

I don't have familiarity with https://github.com/pegjs/pegjs/wiki, one question that comes to mind is regarding tokens that you currently have hardcoded in `grammar.pegjs`. Will you be able to modify this to use the settings contained in:
image

@4Source and @st3v3nmw given the significance of the changes in this PR, would be great if you could both look at this as well.

Cheers
Ronny

alberti42:
> Hi @alberti42 thanks for this, many people have been after this for a long time, and will be very happy when it is released.
> 
> A few requests:
> 
>     1. Currently some of the existing unit test cases fail. Please update your enhancement so that the existing tests all pass. See https://www.stephenmwangi.com/obsidian-spaced-repetition/contributing/#code for details
> 
>     2. Update the user facing documentation, see https://www.stephenmwangi.com/obsidian-spaced-repetition/contributing/#documentation
> 
>     3. Let me know and I'll review at that time
> 
>     4. We will need some new unit test cases. You may prefer to wait till I complete step 3. and provide feedback about functionality.
> 
> 
> Thanks once again! Cheers

It required much more work than what I thought because I wanted to have a clean solution. I created a true parser. It is quick and easy to maintain for the future. The old parser was quite untransparent and could not be easily extended.

The new parser makes use of Peggy. The rules are defined with `peggy.pegjs`. You can find highlighter to supper in your editor highlighting of `peg.js` syntax.

If you modify `peggy.pegjs`, you need to run `npm run grammar` to generate automatically `peggy.mjs` file, which is the real parser.

The `peggy.pegjs` can be tested with https://peggyjs.org/online.html, which is great when you want to test the different grammar rules.

The only part that is not nice in `peggy.pegjs` is the block called `HACK`. You will see it. These are a bunch of functions adapted from the automatically generated `peggy.mjs` files. The purpose is to allow the user-defined keywords, when the user wants to redefine "??" or "?".

For the `multilineCardEndMarker` separating multiline notes, I like to use '---'.

If you leave the field empty in Obsidian configuration pane, then the old behavior will apply.

alberti42:
> Hi @alberti42, I've just noticed that you have rearchitected the parser from the ground up. You must have put in much time and thought into this!
> 
> @4Source initiated the discussion [Rework the flashcard seperator/detecting system #1018](https://github.com/st3v3nmw/obsidian-spaced-repetition/discussions/1018) noting that there are many outstanding requests that relate to the parser and that it's not very easy to enhance in its current form. (The list including the issue titles at https://github.com/users/ronzulu/projects/1)
> 
> So your rearchitect of this part of the plugin might be perfect!
> 
> I don't have familiarity with https://github.com/pegjs/pegjs/wiki, one question that comes to mind is regarding tokens that you currently have hardcoded in `grammar.pegjs`. Will you be able to modify this to use the settings contained in: image
> 
> @4Source and @st3v3nmw given the significance of the changes in this PR, would be great if you could both look at this as well.
> 
> Cheers Ronny

Sorry, I did not see this message when I wrote the previous one just a moment ago.

Yes, we have a new parser, which should be much easier to understand and maintain. The rules can still be simplified a tiny bit to make the grammar more streamlined.. but let's first see what you think of it.

> I don't have familiarity with https://github.com/pegjs/pegjs/wiki, one question that comes to mind is regarding tokens that you currently have hardcoded in grammar.pegjs. Will you be able to modify this to use the settings contained in:

It now works with the user-based keywords. For this, I required what I called a HACK in the previous message. The HACK is pretty straightforward and does not compromise efficiency. At least, it does not make it slower. However, cosmetically, it does not look nice if you are a purist of good coding.

4Source (commented):
I can' say much about the parser it self haven't investigated in it jet. But what I have seen/understand looks great to me for future adjustments and planed features from the discussion

4Source:
> For the `multilineCardEndMarker` separating multiline notes, I like to use '---'.

I think this is a bad Idea we had already issues with the ``---`` because it is used in other formatting like tables or horizontal lines 

> If you modify `peggy.pegjs`, you need to run `npm run grammar` to generate automatically `peggy.mjs` file, which is the real parser.

If I understand this correctly I think how this works should be well documented somewhere in dev guide

4Source:
> Sure. I did not say that this needs to be the default behavior. This is just what I find nice for me. From the preference pane, the user can choose `---` or `+++` or any other sequence of characters she/he likes.

I just meant this should not be the default setting. So user who try this plugin for first time or user there not investigate in the settings have a normally functional experience

alberti42:
> > Sure. I did not say that this needs to be the default behavior. This is just what I find nice for me. From the preference pane, the user can choose `---` or `+++` or any other sequence of characters she/he likes.
> 
> I just meant this should not be the default setting. So user who try this plugin for first time or user there not investigate in the settings have a normally functional experience

I agree 100%. Actually, I would have left the field as empty as the default setting. When empty, the old behavior applies, which is what probably people in the community expect.

ronzulu:
That's great!

@alberti42 , there may be an alternative to the HACK. Have you come across this? Does this help?
[ Dynamically generate / parse rules based on configuration #512 ](https://github.com/pegjs/pegjs/issues/512)

Cheers
Ronny

alberti42:
> That's great!
> 
> @alberti42 , there may be an alternative to the HACK. Have you come across this? Does this help? [ Dynamically generate / parse rules based on configuration #512 ](https://github.com/pegjs/pegjs/issues/512)
> 
> Cheers Ronny

Hey Ronny,

Thanks! Yes, I had read the page before. You can indeed use predicates `& { predicate }` to evaluate whether an expression finally matches or not. This is also nicely documented in https://peggyjs.org/documentation.html.

I have used such predicates already for evaluating dynamically the flags deciding what sybols define a close. They work nicely.

The reason why I did not use it for the other symbols such as `?` and `??` is because I thought the parsing efficiency may be affected. But we can give it a try.

To try this, we would need to define our own grammar of what are the sequences of characters the user is allowed to type. In principle, we could decide any characters except newlines. However, my guess is that the predication function will be evaluated a lot of times to decide if there is a match or not.

There are also other options if you do not like the HACK (I don't like it either).

We could generate the parser.mjs anew directly in Obsidian every time the user changes the configuration in the setting pnae. This happens only rarely anyway, and peggyjs is quite fast to generate from the grammar.pegjs a new .mjs file. The downside is that we need to ship with the app the entire transpiler of peggy.js. So, I would avoid it if we find other and better solutions.

The other possibility is that we use in the grammar.pegjs different PLACEHOLDERSTRINGS123 for each of the strings to be then replaced by the user.

We then write a little .js program of a few lines introducing a second-step transpiler. The purpos is to process the parser.mjs generated by peggyjs. This step can be combined with `npm run grammar`. So it will look like as a single step for the developer of other features.

Such second-step transpiler replaces the static placeholders with javascript code importing the values for each placeholder from the variable `options`. In fact, you may have noticed that it is already possible to dynamically provide content to peggyjs parser, and we are already using it.

This last approach is possibly the most elegant, but not without caveats. If we use static placeholders, peggyjs optimizes the parsing by detecting their string length. Our transpiler would have to replace in the .mjs code the static length with one dynamically optained from the user variables. This is quite efficient in terms of execution speed. However, the transpiler is a bit fragile because it needs to do a surgery on the .mjs file about the length of the PLACEHOLDERS. If something changes in future versions of peggyjs, these changes may break the compatibility with our transpiler, and some small fix will be needed.

Other possibilities. I know it is possible to create plugins for peggyjs, but I do not know if these allow to have dynamical keywords.

And the last possibility, we improve peggyjs and make a pull request to introduce the feature of dynamical keywords.  😅

Bottom line:

If you are happy with the parser, I would focus on testing all features, removing any possible small bugs, and streamlining the grammar file. It can be cleaned up a little. We can then release with the community.

We could think how to replace the HACK in the future. In my assessment, the HACK has no impact on the execution speed or anything else. It is not elegant, and it is not easy to understand someone just opens the code for the first time, and was not given any explanation.

Best,
Andrea

ronzulu:
Hi @alberti42 thanks for such a comprehensive post!

If I understand correctly, the best way would be for the parser to be generated whenever any of the relevant user options are changed. The downside is needing to include the peggy package (and it's dependencies) increasing memory requirements.

Out of curiosity, I tried the following code which dynamically generates the parser:
```js
test("hello", () => {
    var peg = require("peggy");
    var inlineMark = '"::" / "--"';
    var grammar = `
start 
    = q:question inline_mark a:answer { return "Q: " + q +  ", A: " + a}

question = text

answer = text

text = t:[a-zA-Z ]+ { return t.join(""); }

inline_mark = ${inlineMark}
`
    var parser = peg.generate(grammar);

    var actual = parser.parse("What should I eat for dinner::chocolate cake");
    expect(actual).toEqual("Q: What should I eat for dinner, A: chocolate cake");

    actual = parser.parse("What should I eat for dinner--chocolate cake");
    expect(actual).toEqual("Q: What should I eat for dinner, A: chocolate cake");
});
```

I think this is a very clean solution. In terms of package size, I came across this web site:
https://bundlephobia.com/package/peggy@4.0.3

It reported:
image

To me, the benefit outweighs the additional memory usage.

I'm wondering if you have had a chance to look at some of the other related PRs and issues:
https://github.com/users/ronzulu/projects/1

With some of those, there will be more user options, and so whatever we come up with needs to handle the future requirements.

Cheers
Ronny

alberti42:
I tested your example and worked very nicely. Great that it is so smooth to generate the parser from the grammar.

>     var parser = peg.generate(grammar);

I agree, speed and memory consumption should not be a problem. Even if this operation takes a bunch of ms, it is only executed rarely, when the user changes the keywords in the configuration pane.

I would only add a debouncing solution, so the function is not called to many times, each time that the user press a key in the editing field; see for example https://github.com/alberti42/obsidian-plugins-annotations/blob/main/src/main.ts#L566

> I'm wondering if you have had a chance to look at some of the other related PRs and issues: https://github.com/users/ronzulu/projects/1

Not really. Is there anything in particular you think it is important to consider? I think the parser is very powerful and I cannot imagine a situation that cannot be handled.

I was surprised that currently all parsed information is not really stored. At the end, you only store the beginning and end position of the card. In principle, one can make more use of the parser and store the different parts of a flash card in different fields, with annotated the position in the text. But, it seems you did not need this extra structuring of parsed information... so maybe we do not need to put extra work on something like this that is not really needed.

ragrawal:
Any idea when this PR will be merged ?

ronzulu:
> I tested your example and worked very nicely. Great that it is so smooth to generate the parser from the grammar.

That all sounds great. Are you planning to update your code to use this technique?

> Not really. Is there anything in particular you think it is important to consider? I think the parser is very powerful and I cannot imagine a situation that cannot be handled.

I'm sure you're right, so let's not worry about future changes right now.

Thanks!

alberti42:
> That all sounds great. Are you planning to update your code to use this technique?

I have implemented it. It worked for my usage cases. There are some tests that did not pass yet. I will have to look at them in the next days to understand what goes wrong. Hopefully, it will be a very quick fix.

I tested the time to generate the parser. It is already in the debouncing function, which avoids generating the parser too many times if the user changes the options very quickly.

In any case, it takes on my laptop 20 ms. Which is superfast, given the fact that it is generated only once when the plugin is loaded and every time the user changes one of the relevant settings, requiring producing a new parser. So very seldom.

alberti42:
The other day I was in a hurry and had to leave quickly. I checked the messages generated in the test again. It seems to me that all tests are passed. There is some red warning because of some part of code that are never reached. This concerns code when using `try` /  `catch` construct. The catch part is only reached and executed if an error is produced. One would need to include in the test modules some tests generating on purpose the error to execute the code in the catch part. Or alternatively, we can remove the try+catch construct (which is not good style either).

So I think it is not so important, and it is very time consuming to write test modules that verify all try and catch constructs. Did you do so before? Or simply avoided using try/catch constructs?

I show you a screenshot of what the result of the tests is:

Bildschirmfoto 2024-08-15 um 12 57 18

Will you handle the integration of the pull request into the main branch? Is there anything else you want me to do?

alberti42:
Ok. I fetched the upstream changes, and made a single [commit](https://github.com/st3v3nmw/obsidian-spaced-repetition/pull/1012/commits/e2f6750f75456154ba8a05726541cf7043677105) with all changes.

alberti42:
Ok. I understood that I need to do some more changes on the test routines. Also my latest commit does not include some changes I did before which I need to reintroduce. I need some more time to finish it. It should not last long.

alberti42:
I made one more commit. I believe everything works. However, many tests are still not passing because the test routines need to be updated to integrate with the new syntax of the parser.

I am afraid I don't have anymore time to invest in this project.

Regarding the test modules, it is often very difficult for me to understand the reasons of certain rules that seem to be inserted ad hoc. For example, I had to debug a test in `tests/unit/FlashcardReviewSequencer.test.ts` saying:

```
test("Question starts line after tag (single space after tag before newline); New card", async () => {
                let originalQuestionStr: string = `#flashcards${space}
Q2
?
A2`;

                let fileText: string = `
${originalQuestionStr}

#flashcards Q1::A1

#flashcards Q3::A3`;

                let updatedQuestionText: string = `Multiline question
Question starting immediately after tag
?
```

It was not clear to me at all why if I give:

```
#flashcards<space here>
Q2
?
A2

#flashcards Q1::A1

#flashcards Q3::A3
```

then it expects something without the space after `#flashcards`. Why to make such a big deal for the space? And why to remove it? Why does it bother? It is hard to understand these rules and generalize them without creating ad hoc solutions. For the case presented above, I created an ad hoc rule recognizing the tags in a multiline card. But it seems to me not the right approach. For sure I am missing something else.

Also, ideally you would like to parse Obsidian tags with the proper parser. I already made a rule in the grammar.

And the frontmatter could be parsed as well with peggy rules without creating any problem. It does not need to be treated separately.

I need your support. We should reduce the test cases and streamline the code. I myself cannot invest any more time in this plugin. I am sorry.

In the worst case, I will create a new plugin for spaced repetitions from scratch with a shorter and cleaner code.

4Source:
I try to help you with writing and fixing some of the test cases I hope I can push to this pr, if not I will create one on your fork

src/parser.ts
```diff
@@ -1,132 +1,83 @@
...
+    // Use this function when you call the parse method
+	try {
+		cards = parser.parse(text + "\n\n\n",  {
+			CardType,
+			createParsedQuestionInfo: (cardType: CardType, text: string, firstLineNum: number, lastLineNum: number) => {
+				return new ParsedQuestionInfo(cardType, text, firstLineNum, lastLineNum);
+			}
+		});
+	} catch (error) {
+		if (error instanceof SyntaxError) {
```
4Source, line 75: Why did you add the SyntaxError here where could this come from?

alberti42:
Ok final attempt. 

Bildschirmfoto 2024-08-16 um 17 48 04

All tests now passed successfully.

I only modified two tests.

One testing for multicards of basic type with no content in the question. My parser does not detect this type of cards. But I argue, it makes sense. Because what should one reply to a question containing no text? I believe it makes no sense to have such empty questions in a deck. For this test, I added "some content" in the question to make it pass.

The other test I modified was actually disabled. It tests for tags "#flashcard " and checks that the space is removed. I don't understand why we need that. I want to avoid ad hoc patches without understanding the meaning. I presume this is a left over from the previous parser, where all lines added to the cards were treated with "trimEnd()" to remove trailing empty spaces. To implement this with the new parser, it requires quite some extra work, and I would need to understand what the purpose is.

So, from my point of view, the pull request is now ready. You can test it and, hopefully, you will find it useful.

alberti42:
> I try to help you with writing and fixing some of the test cases I hope I can push to this pr, if not I will create one on your fork

Hi @4Source, it would be great if you can have a look. I did not notice your reply and worked more on fixing all test cases, with the exception of two, which I explained above in the previous post. Hopefully, this simplifies your work because in my view, everything works now, if you accept the two test cases I have modified.

src/parser.ts
alberti42: > Why did you add the SyntaxError here where could this come from? Where and when gets this thrown?

You are right. There is no chance that this SyntaxError is generated because we wrote robust parsing rules. Whatever the user types in, the parser will produce a valid end state.

Please feel free to remove it. It is really never triggered.

4Source: ok just got your messages displayed jet I will remove the special case is easier for testing if we can't trigger an syntax error jet

4Source:
> The other test I modified was actually disabled. It tests for tags "#flashcard " and checks that the space is removed. I don't understand why we need that. I want to avoid ad hoc patches without understanding the meaning. I presume this is a left over from the previous parser, where all lines added to the cards were treated with "trimEnd()" to remove trailing empty spaces. To implement this with the new parser, it requires quite some extra work, and I would need to understand what the purpose is.

I think your correct I mean to remember something like this but @ronzulu may now this better I think he wrote the test. If it is like that we should remove it completele

ronzulu:
Thanks for the detailed and comprehensive thinking about this. I've read through all the recent posts and hopefully will be able to reply properly over the next few days.

And totally understand @alberti42 that you can't spend any more time on this. Thankfully it seems like there is nothing further for you to do. If I have specific questions before finalising, hopefully you'll have a small amount of time to answer.

I think what you've added allows us to extend what syntax is supported, without pulling our hair out!

Cheers
Ronny

alberti42:
Thank you @4Source and @ronzulu for your comments and review of the PR, and understanding about my time constraints.

Of course, I am happy to still help and support the project. It would be great to see changes integrated in the upstream.

> In my opinion this could be a case for syntax error. I personal don't like the solution adding "some text" because this was the test case to cover such problems. I think it would be better to change the expected result or treat it as syntax error but than all files starting with a separator would be seen as syntax error

Let me add some explanation about the parser and what a syntax error is.

The important part is:

```
main
  = blocks:block* { return filterBlocks(blocks); }

block
  = html_comment / inline_rev_card / inline_card / multiline_rev_card / multiline_card / close_card / loose_line
```

One way to trigger a syntax error is by removing `loose_line` in the rule `block` above. If there are loose lines [meaning lines added by the user that do not belong to any card (e.g., some comments in between cards, a title, or an obsidian tag)] then the parser would not know, without `loose_line` how to proceed to parse those lines.

Syntax errors by the parser are very bad and you really want to avoid them. If we are parsing a programming language (not our case, but just to make an example), it is clear that a syntax error can be generated when the user writes something that makes no sense or contains some syntax error. Then parser would inform the user that it does not know how to interpret the instructions.

In our case, we do not want to throw errors at the users because they just simply write something beteen two cards, or they use a wrong symbol, for example instead of "??" the user types "???". In my view, it would be bad to display a syntax error. Moreover, when a syntax error is triggered, the parser would stop. Meaning that nothing is parsed at all, it is just an invalid result.

In our case we do not want this to happen. If now the user mistypes "???" instead of "??" nothing bad happens. The parser moves through the card interpreting it as normal `loose_lines`. Of course, the user will notice at some point that the card is not displayed in the deck and will figure out she/he mistyped.

Regarding the test case for a multiline card with a question set to an empty string "", for me it makes no sense to parse it and add it to the deck. With the present grammar, the parser moves through and ignore this block interpreting it as `loose_lines`.

If you really insist that the case of an empty block in front of "?" should be accepted as a possible quesiton, we could look for ways to change the parser to accept it. But I personally do not see any compelling reason for it and why we should invest an extra effort on something that is debatable.

So, if you want to remove the try and catch statement, please feel free to do so. If you want to add a test that triggers the catch, you need to modify the code to allow overwriting the grammar with a custom (weaker) grammar that can produce a syntax error. With the present grammar, I cannot think of any possible text that is not parsed to the end. A syntax error is only produced when the parser encounters some text and does not know how to interpret it and thus cannot reach the end of the note/text.

4Source:
> If you really insist that the case of an empty block in front of "?" should be accepted as a possible quesiton, we could look for ways to change the parser to accept it. But I personally do not see any compelling reason for it and why we should invest an extra effort on something that is debatable.

I think we should change the test case and see it as an error case where no card is detected at all. 

> So, if you want to remove the try and catch statement, please feel free to do so. If you want to add a test that triggers the catch, you need to modify the code to allow overwriting the grammar with a custom (weaker) grammar that can produce a syntax error. With the present grammar, I cannot think of any possible text that is not parsed to the end. A syntax error is only produced when the parser encounters some text and does not know how to interpret it and thus cannot reach the end of the note/text.

@alberti42 I created a pull request to your fork with some small changes I think there would be nice to have. So @ronzulu  please don't rush this pr and merge it before those are in.

alberti42:
@4Source Thank you for the changes and improvements. I took all of them. I did some more commits:

- Added tests for multiline clozes with `multilineCardEndMarker`
- Added more comments to the code
- Improve parser debug infos
- Added a few test cases for testing `multilineCardEndMarker`
- Minor changes to the language translations

In my view, we are ready for the changes to be included in the upstream. What do you think?

Minor question:

Bildschirmfoto 2024-08-17 um 15 34 19

All tests passed successfully. However, some code is not executed. I checked the lines of code not executed, and in some cases it seems to me ok that they are not executed because they concern situations that are not triggered with the tests.

4Source:
> All tests passed successfully. However, some code is not executed. I checked the lines of code not executed, and in some cases it seems to me ok that they are not executed because they concern situations that are not triggered with the tests.

Have you fixed the coverage because for me everything is covered? And I think it would be nice to reach the current conditions because there is an different pr I am working on that improves the overall test coverage

alberti42:
> > All tests passed successfully. However, some code is not executed. I checked the lines of code not executed, and in some cases it seems to me ok that they are not executed because they concern situations that are not triggered with the tests.
> 
> Have you fixed the coverage because for me everything is covered? And I think it would be nice to reach the current conditions because there is an different pr I am working on that improves the overall test coverage

I am not sure I understand your question. I did not change anything in `src/lang/helper.ts` nor in `scheduling.ts`. I did some minor modifications to the language translation, like I added the translation for the Italian one.

I also did some other improvements to make the code more readable and better, as I posted before. And also added more tests about the new added feature about multiline notes, which were not included yet. So, I think these are still changes you want to have.

So, I am not sure why it is now yellow instead of green. I was hoping you would know what I should pay attention to. I agree that it gives a nice feeling when everything is green.

4Source:
I was confused because I checked out your branch and for me everything is green 
image
I thought you may changed something since the screenshot.

> So, I am not sure why it is now yellow instead of green. I was hoping you would know what I should pay attention to. I agree that it gives a nice feeling when everything is green.

And I am pretty sure the workflow will insist on it that the global threshold coverage is 100%

alberti42:
Thanks a lot for the answer. I was confused too. I tried all tests again and now everything is green. Nothing has changed since when I generated the previous screenshot with yellow warnings. Maybe some cache in the terminal session was causing it (but I would not put my money on it)... anyhow, it seems not important now.

I am glad that we have everything green and passed!

alberti42:
@4Source Some idea for the future. If would be nice to have a toggle that displays what lines are parsed as a flashcard. We could use light green for the background. In this way, if the user mistyped something (e.g., typed ??? instead of ??), then he/she would notice it.

I know how to do it from the point of view of the parser to retrieve this information. But I would have to figure out how to modify the visualization of the flashcards in Obsidian editor. I know it can be done. Alternatively (and a bit more elegant), one could highlight the parsed flashcards in the gutter on the left-hand side of the editor.

4Source:
> @4Source Some idea for the future. If would be nice to have a toggle that displays what lines are parsed as a flashcard. We could use light green for the background. In this way, if the user mistyped something (e.g., typed ??? instead of ??), then he/she would notice it.
> 
> I know how to do it from the point of view of the parser to retrieve this information. But I would have to figure out how to modify the visualization of the flashcards in Obsidian editor. I know it can be done. Alternatively (and a bit more elegant), one could highlight the parsed flashcards in the gutter on the left-hand side of the editor.

I like this Idea to give the user a visual feedback there flashcards are working properly. Maybe we could also add a command to display the test for current file.  

> If we want to discuss this seriously, we should open a different channel.

But yes definitely new channel

ronzulu:
> In any case, it takes on my laptop 20 ms. Which is superfast, given the fact that it is generated only once when the plugin is loaded and every time the user changes one of the relevant settings, requiring producing a new parser. So very seldom.

Great! Thanks for testing the performance.

> There is some red warning because of some part of code that are never reached. This concerns code when using try / catch construct. 
>S o I think it is not so important, and it is very time consuming to write test modules that verify all try and catch constructs. Did you do so before? Or simply avoided using try/catch constructs?

I agree, sometimes it's too difficult and not worth the effort to get 100% test coverage.


> Will you handle the integration of the pull request into the main branch? Is there anything else you want me to do?

I'm only planning to complete the review, and I'll leave merging into the main branch to @st3v3nmw .

> then it expects something without the space after #flashcards. Why to make such a big deal for the space? And why to remove it? Why does it bother? It is hard to understand these rules and generalize them without creating ad hoc solutions.

All valid comments. Honestly I don't remember the reason for the test, but presumably there was one. And in this case of a line just containing a tag, there is no reason to care whether or not the trailing space character is removed. I think here it's just a case that the `checkUpdateCurrentQuestionText` method does an exact string match of the result.

I've deleted the test case that's been commented out.

> Also, ideally you would like to parse Obsidian tags with the proper parser. I already made a rule in the grammar.

Just checking… The "tag" rule is defined, but not used at present. Is that correct?

> And the frontmatter could be parsed as well with peggy rules without creating any problem. It does not need to be treated separately.

Good idea, that would be cleaner. But I'm sure many other priorities…

> In the worst case, I will create a new plugin for spaced repetitions from scratch with a shorter and cleaner code.

Do you still think that might be a possibility? Hopefully not!

> Ok, maybe the only usage case of this try-catch statement is for future modifications. If someone does modifications to the rules, it is possible that the parser is not able to parse the whole text without throwing a parse syntax error. But I leave it to you whether to include this try-catch for future possible pitfalls or not. Now we do not need it.

I've added a comment regarding syntax errors into the code.

> One testing for multicards of basic type with no content in the question. My parser does not detect this type of cards. But I argue, it makes sense. Because what should one reply to a question containing no text? I believe it makes no sense to have such empty questions in a deck. For this test, I added "some content" in the question to make it pass.

The test description is "Multi-line without question (i.e. question is blank)" - so I've updated the test case to check that only Q2/A2 is parsed

```
---
?
A1

Q2
?
A2
```

> Some idea for the future. If would be nice to have a toggle that displays what lines are parsed as a flashcard. We could use light green for the background. In this way, if the user mistyped something (e.g., typed ??? instead of ??), then he/she would notice it.

Agreed, let's discuss separately…

### Other

I've made some minor naming changes (hopefully makes easier to understand)

`not_newline`->`text_till_newline`
`_`->`whitespace`
`_*`->`optional_whitespace`
`[^\\n]` -> `non_newline`

###  PR

> @alberti42 I created a pull request to your fork with some small changes I think there would be nice to have. So @ronzulu please don't rush this pr and merge it before those are in.

I'll try to do the same

ronzulu:
Hi @alberti42 

Sorry, but I can't work out how to create the pull request.

Are you able to merge the changes from my branch?
https://github.com/ronzulu/obsidian-spaced-repetition/tree/rz-alberti42

Note that I created that branch from upstream master, and not from your repository.

If too difficult, let me know and I'll simply post the small number of files.

Cheers
Ronny

alberti42:
> Do you still think that might be a possibility? Hopefully not!

No, I prefer to be helpful with the existing plugin. Some of the code of obsidian spaced repetition is not very stylish, but it gets the job for the user done very well!

> Just checking… The "tag" rule is defined, but not used at present. Is that correct?

Yes, it was an attempt to deal with the rule/test we have discussed. It worked, but I felt that the rule/test was not well justified. So, I removed having to deal with tags in the current parser, because it was really ad hoc only for that rule/test.

However, I did not delete it because we may need it one day if we want to expand the parser to properly detect tags and frontmatter.

Also, we already use the parser to identify the different parts of a flashcard. At the moment, we detect all parts, but then throw away this information. It would be much nicer for the code if we store infos about the parsed cards, and then use it later without reparsing the individual cards when using them.

alberti42:
> Are you able to merge the changes from my branch? https://github.com/ronzulu/obsidian-spaced-repetition/tree/rz-alberti42

Yes, I have done it. However, the tests now throw hundreds of errors. I checked again. Before, it was 100% pass everywhere.

alberti42:
> I've made some minor naming changes (hopefully makes easier to understand)

Yes, I have seen them and agree with the improvements.

The errors in the tests come certainly from other changes you made that do not concern my PR. I simply accepted all your changes in your branch and did not filter those changes regarding my PR only.

ronzulu:
> The errors in the tests come certainly from other changes you made that do not concern my PR. I simply accepted all your changes in your branch and did not filter those changes regarding my PR only.

Sorry, that's all my fault... this is the first project that I've worked on using git. I think the main problem is that the "squashed commit" didn't track the connection to your branch from where the changes were obtained.

I'll do this cleanly, creating a new fresh branch starting from your "Minor cleanup of the parser grammar"

image

ronzulu:
Hi @alberti42 I've published the new branch https://github.com/ronzulu/obsidian-spaced-repetition/tree/rz-alberti42-v2.

I'm updating the user documentation as well, I'll commit that separately.

ronzulu:
Hi @alberti42 

As I'm writing up the user documentation, I noticed that your enhancement permits blank lines in the answer of a multiline flash card, but not in the question. Is this intentional?

You've put in so much work into this, what do you prefer?
1. Release the PR as is, and the modification can be sorted out down the track
2. We make the change now

A complexity is that there is slightly different behaviour if a blank line is the delimiter or if there are actual characters. No blank line would be needed at the start of the file (or immediately after the front matter)

Cheers
Ronny

ragrawal:
I have been closely following this PR mainly because I am super excited about having multiline support . So throwing out my opinion. I feel multiline support is more important than in answer than questions. So I don’t want to delay this PR and would love it to see go out. You can always add multiline support in question as an additional capability. Just my two cents 😬

4Source:
> I have been closely following this PR mainly because I am super excited about having multiline support . So throwing out my opinion. I feel multiline support is more important in answer than questions. So I don’t want to delay this PR and would love it to see go out. You can always add multi line support in question as an additional capability. Just my two cents 😬

@ragrawal We already have multi line support.

``` md
<!-- empty line here is important -->
This is a multi line example in <!-- no empty line here -->
the question 
? <!-- must be in separate line -->
and in the <!-- no empty line here -->
answer
<!-- empty line here is important -->
```

the ``?`` is default but you can double check it in the settings ``Seperator for multiline flashcards``

image

alberti42:
> Sorry, that's all my fault... this is the first project that I've worked on using git.

Hi @ronzulu, you can use ChatGPT. It is quite helpful to give you precise instructions what to do from command line if you describe your problem. Git and GitHub support a myriad of possibilities (which is hardly possible to remember and know all).

I am not sure where I have to follow updates. Will you post here in this channel an update when the PR will be finally integrated upstream?

alberti42:
> Hi @alberti42
> 
> As I'm writing up the user documentation, I noticed that your enhancement permits blank lines in the answer of a multiline flash card, but not in the question. Is this intentional?
> 
> You've put in so much work into this, what do you prefer?
> 
>     1. Release the PR as is, and the modification can be sorted out down the track
> 
>     2. We make the change now
> 
> 
> A complexity is that there is slightly different behaviour if a blank line is the delimiter or if there are actual characters. No blank line would be needed at the start of the file (or immediately after the front matter)
> 
> Cheers Ronny

Hi Ronny,

It was intentional from my side because I did not need to make questions spanning over multiple lines. We can think about what a convenient way is to denote the beginning of a question. However, at the moment we do not have a clearly defined symbol where the question is supposed to start. The current implementation checks that we have a block of text without empty lines followed by either "?" and "??".

Of course, we could add in the future a new symbol denoting the beginning of a multiline question containig empty spaces.

One can choose the same characters/symbol to denote the end of a multiline question. In this way, where one question ends, the new one starts.

It kind of make sense to me and can I can imagine it should not be hard to implement. However, I would not delay the release of the new feature for having blank lines in the questions.

Once this PR is released, then we can open a new branch to implement multilines in the question.

That being said, I leave the final decision to you whether you want to have everything together or not. I do not have time in the next days to program this feature but I can imagine in the next 5 weeks it can be done.

ragrawal:
> > I have been closely following this PR mainly because I am super excited about having multiline support . So throwing out my opinion. I feel multiline support is more important in answer than questions. So I don’t want to delay this PR and would love it to see go out. You can always add multi line support in question as an additional capability. Just my two cents 😬
> 
> @ragrawal We already have multi line support.
> 
> ```
> 
> This is a multi line example in 
> the question 
> ? 
> and in the 
> answer
> 
> ```
> 
> the `?` is default but you can double check it in the settings `Seperator for multiline flashcards`
> 
> image

Sorry by multiline, I meant empty line support in answer . Lot of my cards have table and equations. They do tend to have empty lines . Hence waiting for this PR

alberti42:
Hi Ronny, sure, no problem, let me when you are ready and then restore the old branch of mine and repeat the merge with your branch.

4Source:
> (Note: I've never had to restore a branch to a previous state. Hopefully that can be done safely.)

@ronzulu Not sure this helps you but you can [revert](https://git-scm.com/docs/git-revert) a commit. Alternatively you can [reset](https://git-scm.com/docs/git-reset) branch without losing the changes not sure how good this works for already pushed commits
I would recommend first trying to use reset keeps the history cleaner 
``git reset HEAD~<number of commits to go back>`` Keeps the changes as staged 
``git revert <commit>`` creates new commit which undo all back to the commit

ronzulu:
Hi @alberti42 

I think I've found a bug whilst writing the user notes.

Here is a multiline card with a table in the answer...

image
[Table with preceding blank line and+++.md](https://github.com/user-attachments/files/16734994/Table.with.preceding.blank.line.and%2B%2B%2B.md)

This displays correctly during review...

image

image

But after clicking on `hard`, the scheduling info isn't added to the file, and this error is shown in the error log

image
[obsidian.md-1724468486804.log](https://github.com/user-attachments/files/16735007/obsidian.md-1724468486804.log)

Cheers
Ronny

alberti42:
Hi Ronny,

Good that you found the bug. Can you provide the text of your flash card and how you configured the delimiters? In other words, are you using the default settings? Or you chose some special character/symbol?

I will have a look at it.

alberti42:
To my understanding it has nothing to do with the tables.

```
Testing card
?

Line 1
Line 2

---
```

The problem seems to be when you have multicards with empty lines and there is an empty line after `?`. I will check the details.

alberti42:
Hi Ronny,
I found the problem. It was my fault. I used `.trim()` instrad of `.trimEnd()` in the parser. I fixed the problem and added one more test case.

ronzulu (approved):

ronzulu:
Hi @alberti42 I've completed changes to the user guide at:
https://github.com/ronzulu/obsidian-spaced-repetition/tree/rz-alberti42-v2

Feel free to merge and make any changes.

The end is in sight!

alberti42:
Wonderful. I accepted all your changes. All tests are green!!

st3v3nmw:
Thanks @alberti42, @ronzulu & @4Source for the work put in!

ronzulu:
> Thanks @alberti42, @ronzulu & @4Source for the work put in!

I think if this was IRL we would be going to the pub for a drink to celebrate :-)

ragrawal:
So do I just need to reinstall the plugin to get the latest version of the spaced repetition plugin

alberti42:
> I think if this was IRL we would be going to the pub for a drink to celebrate :-)

🎉🎉🎉 Absolutely 😀

alberti42:
> So do I just need to reinstall the plugin to get the latest version of the spaced repetition plugin

I don't think you need to uninstall it. However, you need to wait for the upstream changes to be part of the new release. I am not sure when this will happen, but I celebrate that the changes are now part of the upstream and the release will happen at some point.

ragrawal:
wondering when will this be available. Seems like the updated the plugin was never released. The github page shows the last release to be 1.12.5 which was about a month ago

ronzulu:
This is now part of the latest beta release:
https://github.com/st3v3nmw/obsidian-spaced-repetition/discussions/1011

ronzulu:
> wondering when will this be available. Seems like the updated the plugin was never released. The github page shows the last release to be 1.12.5 which was about a month ago

Hi @ragrawal, wondering if you've had a chance to try this out yet. #1011

ragrawal:
I started using it. Thanks. It’s working great. Didn’t notice any issue .

Issue #839, opened 2024-01-13
[FEAT] For improved UX, split the long list of plug-in options across categories within a tab control


ronzulu:
**Is your feature request related to a problem? Please describe.**

The list of options supported by the plug-in is quite extensive. This list likely will grow when various feature enhancements are implemented, and will become somewhat daunting especially for new users.

**Describe the solution you'd like**
Split the current list by category by using a tab control, such as used by the "shell commands" plug-in

![image](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/75528127/0e0b414a-59c6-4160-90ae-0506f630d0ab)

![image](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/75528127/2d28b8ac-e7f9-406e-a468-923a0bfd9ec5)

Steindvart:
A good idea. It looks amazing! I didn't even know that there is such an opportunity in the organization of the settings interface. I will try to implement it in the near future.

ronzulu:
Sorry @Steindvart I've started without you!

Steindvart:
> Sorry @Steindvart I've started without you!

Thanks! Then we will wait for such a wonderful update :)

In connection with this situation, such an idea arose – it is necessary to mark in the Issue (comment or in description) that you started working on it. This will be a good practice to avoid isolated cross-work on a particular problem.

Steindvart:
Hi @ronzulu! Looks awesome! Thank you for your work!

The compromise with the `Advanced` tab, I think, is justified and looks the best at the moment.

st3v3nmw:
Hi @ronzulu,

I agree with @Steindvart, the UI improvement is amazing! I'm also not too keen on the name "Advanced", I can't think of a better alternative at the moment

How does it look like on mobile?

ronzulu:
Thanks @Steindvart  and @st3v3nmw 

I've had an idea instead of the "Advanced" tab.

Simply include the setting in both the Flashcard and Notes tabs.

This means that we don't have to struggle to come up with some alternative word to "Advanced", and has the benefit that it's directly visible to users where it is most relevant.

I've added some words "Note that this setting is common to both Flashcards and Notes.

![image](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/75528127/17709bf6-bbb0-4ef6-9f0f-3a7291ed7898)

[...]

PR #1021, opened 2024-07-19
[FEAT] Split the long list of options into categories within a tab control

ronzulu:
See https://github.com/st3v3nmw/obsidian-spaced-repetition/issues/839


st3v3nmw (approved):
LGTM, thanks

PR #1027, opened 2024-07-22
feat: extended card title

KyleKlus:
#### This PR implements an extension to the displayed card title partly as described in issue #851.

I added this because sometimes one doesn't know how many cards are left to review and thus is discouraged to continue if the time one has is short, but if the user exactly knows how many cards are left, then one could guess how long it will take and just pull through with it.

Another good reason for this addition is that the card title will display more useful information in the random card mode.

#### Before:

Before
_It just displayed the current deck name and the number of cards left in it._

#### After (with random card random deck mode enabled):

After_random_mode
_In random mode it displays now the deck chosen, with the number of cards left and in the subtitle the deck out of which the card was drawn._

#### After (with sequential mode enabled):

After_sequential _mode
_In sequential mode the title will display the deck chosen, with the number of cards left and in the subtitle the number of subdecks, the current deck name and the number of cards left in it._


st3v3nmw (approved):
LGTM, thanks.

Please have a look at the merge conflicts.. I can't resolve them on the UI.

KyleKlus:
Hi, everything is now resolved.

Issue #1033, opened 2024-08-01
[Bug]: After update 1.12.5 does not work in the mobile version of obsidian (android)


antomal:
### Describe the bug

Just updated the plugin to version 1.12.5 and now I can't open the cards and the queue selection window! :-(

### To Reproduce

1. Update plugin version to 1.12.5
2. Try to open the card list or queue

### Expected behavior

I should see a window with lists of cards or a window with tags on pages (queue)

### OS of your device

MIUI 13.0.6

### Obsidian version

1.6.7 (149)

### Plugin version

1.12.5

ronzulu:
Hi @antomal 

We'll need a more comprehensive description of what fails, including screenshots.

It seems fine on my Samsung Android

Cheers
Ronny

![2024-08-01 20 58 25](https://github.com/user-attachments/assets/4d93dc1d-6d39-4c77-bd1a-c1e6441ed6f7)
![2024-08-01 20 58 39](https://github.com/user-attachments/assets/b3e19ece-25d2-4d6d-a930-aea0f8e1dab7)
![2024-08-01 20 58 46](https://github.com/user-attachments/assets/0bbd91ad-cbb6-46db-b062-b6b1fabba25f)
![2024-08-01 20 58 53](https://github.com/user-attachments/assets/9abcdf03-cb93-4f51-a9bc-c7eeb457237b)
![2024-08-01 20 59 01](https://github.com/user-attachments/assets/e339513b-329a-4b5b-b6e9-ed9d03a5a12b)

antomal:
I'm happy to help, but I don't know how to describe in more detail and make screenshots of what is not there.

But I'll try. I recorded a video from the screen.
https://static.antomal.com/Screenrecorder-2024-08-01-14-17-27-933.mp4

1. first I show that the version is 1.12.4 and everything works
2. I update the plugin to 1.12.5
3. I try to open window "Review flashcards" - nothing happens
4. I try to open command "Spaced Repetition: Open a note for review" - nothing...

ronzulu:
Sorry, I don't have much expertise on an android platform.

If this was on the desktop, I'd suggest opening the chrome developer window (control shift I) and see if there is any error output.

I did a Google search, but couldn't find an easy way to do this on Android. One way that involves connecting the phone to a PC is shown here (but I haven't tried it)
https://help.pushwoosh.com/hc/en-us/articles/360027726731-Android-Chrome-console-logs

[...]

PR #1032, opened 2024-07-30
Major reworking of the documentation

ronzulu:
Fixes #1029,  #876

Available for reviewing at:
https://rjz-osr.s3.ap-southeast-2.amazonaws.com/site/en/index.html

PR #1048, opened 2024-08-12
[FIX] isEqualOrSubPath function

4Source:
Added better description discussed in #972
Stopp using the normalize function discussed in #1033


ronzulu (changes requested):
Hi @4Source LGTM, could you update the change log. Thanks!

4Source:
@ronzulu what I am doing wrong? I run ``npm run changelog``

ronzulu:
Hi @4Source.

There are actually a few different issues here:

1. How updates to the change log should be done for a PR
2. Incorrect updating of references within [Major reworking of the documentation #1032](https://github.com/st3v3nmw/obsidian-spaced-repetition/pull/1032)(my fault)
3. Auto generation of the change log

# Updating the change log in a PR

I don't think we can use auto generation of the change log whilst working on a PR as it generates the change log based on what has already been merged into master (I believe).

So I've always just manually updated the change log as described in "Add your change to the [Unreleased] section of the changelog" in https://www.stephenmwangi.com/obsidian-spaced-repetition/contributing/#code.

So I think that's probably what needs to happen now with your PR.

# Incorrect updating of references (my fault)

Background

With the major reworking of the documentation https://github.com/st3v3nmw/obsidian-spaced-repetition/pull/1032, I created a test vault for taking screenshots for the user documentation.

I tried to create this under the existing directory structure, i.e. under `/docs`. But I can find a way to stop `mkdocs` from including those files within the generated static website.

So my solution was to create a new level in the directory structure, as:
image

I tried to change all references, but I didn't catch all, including the reference in file (root level) `/changelog.md`.

I have fixed this now on the documentation branch.

# Auto generation of the change log

Out of curiosity I ran `pnpm run changelog` to update the change log automatically (something I've never needed to do).

For some reason even though `docs/docs/changelog.md` is specified in the command `auto-changelog --template=compact --package && npx prettier --write docs/docs/changelog.md` (`package.json`), the change log was generated at `/changelog.md`.

This seems to also be the case in what happened with you… `/changelog.md` got updated rather than `docs/changelog.md`.

4Source:
> # Auto generation of the change log
> Out of curiosity I ran `pnpm run changelog` to update the change log automatically (something I've never needed to do).
> 
> For some reason even though `docs/docs/changelog.md` is specified in the command `auto-changelog --template=compact --package && npx prettier --write docs/docs/changelog.md` (`package.json`), the change log was generated at `/changelog.md`.
> 
> This seems to also be the case in what happened with you… `/changelog.md` got updated rather than `docs/changelog.md`.

I think this should be possible to auto generate the unreleased section with actions

ronzulu (approved):
Thanks

st3v3nmw:
> This seems to also be the case in what happened with you… `/changelog.md` got updated rather than `docs/changelog.md`.

`changelog.md` is a symbolic link to `docs/changelog.md`. It mostly works okay but gets all messed up on a non-unix OS.

---

Because of the conflicts we've been having with the `changelog.md` file, we can stop updating the file within the PRs and only do it during releases. I'll update the docs..

PR #1057, opened 2024-08-30
add tagsToIgnore setting

Newdea:
FYI, this a bug posted in my branch: https://github.com/open-spaced-repetition/obsidian-spaced-repetition-recall/issues/72
And i think the plugin maybe need it.


st3v3nmw:
Hi @Newdea.. Thanks but I think the better solution would be to let the user use glob patterns to ignore folders and/or files. I've opened a PR -> https://github.com/st3v3nmw/obsidian-spaced-repetition/pull/1086.

Newdea:
yes, using glob patterns is much better, i will close this pr.

PR #947, opened 2024-05-01
Filter due notes when all are scheduled

carlosgaldino:
Ignore notes due in the future.

Fixes #548.

PR #1088, opened 2024-09-23
fix: Review only due notes

### LINKED ISSUE #328: [BUG] `open a note for review` can not open notes correctly.
**Describe the bug**

**To Reproduce**
1. Click `open a note for review` in state bar.
2. The plugin opens the notes that should be reviewed tomorrow.

**Expected behavior**
Because there is no note that should be reviewd today, so the plugin should open a note in `New` queue.

**Screenshots**

https://user-images.githubusercontent.com/47976614/144002560-0cffc5ea-cdcf-4a31-9690-fd053529a644.mp4

 - OS: win11 insider preview build 22504
 - Obsidian version: v0.13.6
 - Plugin version: v1.6.2
 - If on desktop, Installer version: "5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) obsidian/0.12.5 Chrome/89.0.4389.128 Electron/12.0.11 Safari/537.36"

**Additional context**
No

carlosgaldino:
This patch fixes the behaviour of determining the next note to only consider notes that are due. If a note is scheduled in the future, it won't be picked up for review until it's due.

This is the same patch as #947.

Fixes #328.

### DIFF
--- a/src/note-review-deck.ts
+++ b/src/note-review-deck.ts
@@ -1,5 +1,6 @@
 import { t } from "src/lang/helpers";
 import { ISRFile } from "src/sr-file";
+import { globalDateProvider } from "src/utils/dates";
 import { globalRandomNumberProvider } from "src/utils/numbers";
 
 export class SchedNote {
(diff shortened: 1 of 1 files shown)


st3v3nmw (approved):
Hi, thanks for fixing the regression.

carlosgaldino:
Thanks @st3v3nmw. There was a condition missing, I opened #1089 with the change.

PR #1089, opened 2024-09-24
fix: Only consider filtered notes for next note

carlosgaldino:
The previous patch #1088 missed to change the conditional to also include only the filtered notes. With this patch, if all scheduled notes are in the future, it will choose a new note for review.


st3v3nmw (approved):
Thanks!

PR #1095, opened 2024-09-25
fix: remove caret from package manager definition

### LINKED ISSUE #1091: [FEAT]: Specify the package manager (pnpm) without a caret

I can't run **pnpm install**, due to the syntax used to define the package manager. New contributors to this plugin could get stuck if they can't install dependencies.

I get this error if I run pnpm install
image

I will define the package manager and version without the caret(^), which is not valid in the package manager field, like this:
```bash
"packageManager": "pnpm@9.10.0"
```
Additionally, I will add "**engines**" field which will allow us to specify the minimum version of pnpm  required for this project. In this case, pnpm version 9.10.0 or higher are used, 
```bash
"packageManager": "pnpm@9.10.0",
"engines": {
  "pnpm": ">=9.10.0"
},
"engineStrict": true
```

### Additional context

Current versions:
Node: 20.12.1
Pnpm: 9.10.0

EspiraMarvin:
- removes caret from pnpm package manager definition
 - specifiy minimum pnpm version

-fixes #1091

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,5 +52,9 @@
         "peggy": "^4.0.3",
         "short-uuid": "^5.2.0"
     },
-    "packageManager": "^pnpm@9.10.0"
+    "packageManager": "pnpm@^9.10.0",
+    "engines": {
+        "pnpm": ">=9.10.0"
+    },
+    "engineStrict": true
 }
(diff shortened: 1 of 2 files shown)


st3v3nmw (approved):
LGTM, thanks.

PR #1107, opened 2024-10-05
fix: Sort scheduling info and replace values individually (resolves #613)

### LINKED ISSUE #613: [BUG] YAML duplication due re-ordering
**Describe the bug**
Changing YAML list order causes sr-* entry duplication.

**To Reproduce**

Add a new YAML entry at the bottom of the metadata list.

**Expected behavior**
The plugin should replace all sr-* entries by search & replace, not expecting the entries to be at the bottom.

**Additional context**
I am using linter to sort YAML key order, which causes this issue.
Regardless of linting or not, one has to keep in mind to be careful about the metadata entries they're adding otherwise it would cause duplaction of entries for this plugin, which I think could be avoided either with a search & replace, or a completely different approach, as #162.

thisispvb:
- fix: Sort scheduling info and replace values individually
- fix: Ensure consistent timezone when running tests _(otherwise failed locally for me in `GMT-0400`)_

_NOTE: The whitespace changes are from prettier_


src/data-store-algorithm/data-store-in-note-algorithm-osr.ts
```diff
@@ -51,30 +52,43 @@ export class DataStoreInNoteAlgorithmOsr implements IDataStoreAlgorithm {
...
+            );
+
+            const schedulingEase = SCHEDULING_INFO_EASE_REGEX.exec(fileText);
+            fileText = fileText.replace(
+                SCHEDULING_INFO_EASE_REGEX,
+                `${schedulingEase[1]}${ease}${schedulingEase[3]}`,
+            );
+
+            const schedulingInterval = SCHEDULING_INFO_INTERVAL_REGEX.exec(fileText);
             fileText = fileText.replace(
```
st3v3nmw, line 76: I've just checked the Obsidian API and it's now possible to modify the FrontMatter using the native `processFrontMatter` function. That might be the better approach.

https://github.com/obsidianmd/obsidian-api/blob/23947b58d372ea02225324308e31d36b4aa95869/obsidian.d.ts#L1551

>   /**
     * Atomically read, modify, and save the frontmatter of a note.
     * The frontmatter is passed in as a JS object, and should be mutated directly to achieve the desired result.
     *
     * Remember to handle errors thrown by this method.
     *
     * @param file - the file to be modified. Must be a Markdown file.
     * @param fn - a callback function which mutates the frontmatter object synchronously.
     * @param options - write options.
     * @throws YAMLParseError if the YAML parsing fails
     * @throws any errors that your callback function throws
     * @example
     * ```ts
     * app.fileManager.processFrontMatter(file, (frontmatter) => {
     *     frontmatter['key1'] = value;
     *     delete frontmatter['key2'];
     * });
     * ```
     * @public
     */
    processFrontMatter(file: TFile, fn: (frontmatter: any) => void, options?: DataWriteOptions): Promise<void>;

st3v3nmw (changes requested):
Thanks for the pull request. Please check the comment.

Issue #950, opened 2024-05-05
open card in note tab replace floating tab


arias007:
open card in note tab replace floating tab like pkmer.

arias007:
This way, other files and card tables can be quickly opened without waiting for the card table to load or losing progress every time the card is reopened

ronzulu:
Hi @arias007 

This seems similar to [ [FEAT]: leaving modus "review flashcards" and directly jump to note #786 ](https://github.com/st3v3nmw/obsidian-spaced-repetition/issues/786)

If it's similar, then best to add any additional information on that issue rather than here.

If this is different, you will need to explain thoroughly, including screenshots with annotations of what you are wanting.

Cheers
Ronny

arias007:
This can solve many problems, such as
1. We can use the tts plugin to read cards aloud
2. Card review and file review do not need to be one on the left and one on the right. We can concentrate on processing cards or notes, and process them like cards. We do not need to open easy or hard (with the date visible) in the upper left corner, and can directly jump to the original file or note in the new tab through the link. When we return, the card progress remains the same as before, which also solves the problem of opening notes and PDF
4. Shorten unnecessary overloading, each time a tab is opened, it will be the same as the last one opened. If overloading is needed, a new tab can be opened again
five

arias007:
Review cards or notes should be concentrated on one tab instead of floating tabs
like this
![image](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/90596315/226eb80f-135e-4989-aa6a-d0424763ded8)

arias007:
![image](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/90596315/650cc4b7-fde7-477b-a7f6-00e72acab236)

arias007:
![屏幕截图 2024-05-05 100053](https://github.com/st3v3nmw/obsidian-spaced-repetition/assets/90596315/046200a7-fde7-46bf-908f-fff54f439622)

ronzulu:
Hi @arias007, thanks for the additional explanation, that helps.

I think what you are suggesting would also satisfy [ [FEAT] Separate the spaced repetition window or open it in a separate tab within Obsidian #819 ](https://github.com/st3v3nmw/obsidian-spaced-repetition/issues/819)

[...]

Issue #819, opened 2023-12-21
[FEAT] Separate the spaced repetition window or open it in a separate tab within Obsidian


nayyala:
Hey @ronzulu 

As I posted before, I had an issue regarding YouTube playability in Flashcards. It turns out that general YouTube playback for specific timestamps is only supported with Media Extended, which, when in a flashcard, opens a separate tab that cannot be viewed properly because the Flashcard Review panel covers it and blocks access to it without losing the current spot.

If there were an option to separate the window or put it in a tab, this might enable not only Video based flashcards, but also flashcards that might use bigger images. I know that I would like to use bigger images personally, but I find that I have to keep them somewhat limited to fit within the window

KyleKlus:
Hi, this one is implemented now 😊

PR #1169, opened 2024-11-17
feat: Option for viewing sr decks and cards in tabs

KyleKlus:
This pr implements the ability to open the views (deck view and card view) of the sr plugin in a tab/window, rather than in a modal. This ability is turned off by default and can be turned on via a toggle in the ui settings of the sr plugin.
Further more one can also limit the size of the ui within a tab via the already existing flashcard size setting.

This should also satisfy the issues #819, #950 and #1180 partly, but I can also add the open file in new tab button, if that is alright with you, as I've already implemented that on an unpublished branch.

image
_The deck list in a tab_

image
_A card in a tab_

image
_New setting in the "User Interface" tab_


KyleKlus:
Hi, sadly I cant get the import sorting to work, as the formatter doesnt fix any sorting issues for me.

KyleKlus:
I've also discovered that not just my new ui breaks in horizontal mode on mobile, but all the other ui too, so I think, that fixing that should be a different pr, as this one is already getting quite big.

I can take care of fixing that later on in a different pr.

BlitzVI:
hey ik this might be a dumb questions because i do not know how to use github but i'd really like to use this feature but i dont see the toggle in user interface even tho the merge was committed 5 days so can you please help me out? tysm in advance!

KyleKlus:
Hi, the new feature will only be visible in obsidian once a new release version has been published. 

So you could either wait till then or you would have to build a version of the plugin for yourself, as described in the contributing section on the website for this plugin.

BlitzVI:
ah how can i build a version of this plugin using this fork for myself? i'd really like to try this out rn as i think it would be really helpful for my workflow

Issue #1171, opened 2024-11-20
Spaced repetition does not work on Android


aaronstarky:
I really love this plugin and would like to be able to do my flashcards on my phone. unfortunately, the UI is too small to be able to be used on phones. Please help!

MostlyArmless:
Fortunately we can fix this ourselves with some CSS. Create a file in your vault's `.obsidian/snippets/` folder called `fix-flashcard-modal.css` and paste this in:
```css
.modal#sr-modal {
  width: 100vw !important;
  max-width: 100vw !important;
  height: 100vh !important;
  max-height: 100vh !important;
  margin: 0 !important;
  position: fixed !important;
  left: 50% !important;
  top: 50% !important;
  transform: translate(-50%, -50%) !important;
}
```

Then restart Obsidian, go to Obsidian's settings --> Appearance tab, scroll all the way down and turn on the toggle for the CSS file you just added. All the flashcard popup dialogs should now entirely fill the screen.

PR #1177, opened 2024-11-25
fix: modal CSS to make modal full screen

MostlyArmless:
The plugin is unusable on mobile #1171
This is beacuse the modal renders too narrow to be usable on mobile. Fixing this by making the modal fullscreen, across all platforms.

Here's the before and after on desktop:
|Before|After|
|-|-|
|Pasted image 20241125095107|Pasted image 20241125095229|

### DIFF
--- a/styles.css
+++ b/styles.css
@@ -31,10 +31,15 @@
 }
 
 #sr-modal {
-    position: absolute;
-    top: 50%;
-    left: 50%;
-    transform: translate(-50%, -50%);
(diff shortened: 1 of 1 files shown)


mxschll:
You can set the flashcard width and height in the settings, which works perfectly fine for me on mobile.

st3v3nmw:
Also, @KyleKlus's #1194  is close to landing so we'll go with that instead. Thanks for the fix regardless! 😅

PR #998, opened 2024-06-16
FEAT-990 Mobile landscape mode + fully functional resize sliders

annappropriate:
Based on the discussion in https://github.com/st3v3nmw/obsidian-spaced-repetition/discussions/990

- added CSS to make UI elements rotate when the mobile device is rotated
- made size sliders change the `max-width` and `max-height` to make them fully effective (proper full screen when width and height are at 100%)

Tested on desktop and mobile.

PR #1194, opened 2024-12-28
fix: Mobile screen rotation fix

KyleKlus:
This pr fixes the current visual bug, when the plugin is used on a mobile device in landscape mode (See pictures).
Further more this pr also adds a "how to develop & test ui changes" to the contribution page, so that these issues wont arise again in a future ui change.

## PC design

_For reference: Old design used on a pc_

_New design used on a pc_

## Mobile design

_For reference: Old design on a mobile device in landscape mode with the visual bug_

_New design on a mobile device in landscape mode_

_Old design on a small mobile device in landscape mode with visual bug_

_New design on a small mobile device in landscape mode_


KyleKlus:
## The thought process behind the end design

The whole thing broke because I didnt check the mobile landscape mode, when I developed the more detailed session progress title text and didnt assume that someone moved everything around so dramatically.
Someone wanted to make room in the mobile landscape mode and has thus moved all the buttons to the side, to open up some space in the middle (see #990 & #998).

To fix this, I moved the progress text to the card context area, shrunk down the response buttons for small screens and only moved the upper control buttons to the left side in the landscape mode, such that more space is opened up for viewing the cards. I did my best to respect the wish to have more viewing space and at the same time to have a good and consistant design

KyleKlus:
The last commit changes the looks of the deck progress info a tiny bit(see pictures below).

**Changes:**

- _Deck progress info now with bold font weight:_



<br/>
<br/>

- _Chosen deck progress info now without sub deck counter if there are no subdecks:_

bentoner:
@KyleKlus Thanks for fixing this. Works for me on iPad and Desktop.

st3v3nmw (changes requested):
Hi @KyleKlus, thanks for the fix!

My only question is, can we make the _card context_ always appear on a separate line on its own? Something like this:

image

Currently, it moves to the bottom when the _card context_ is too long which is a bit distracting if it keeps switching places:

image

KyleKlus:
Sure, let's do that 😊.

I'll work on it either this evening or tomorrow, so it should be ready until Tuesday.

st3v3nmw (approved):
LGTM, thank you for the UI fixes!

Issue #1175, opened 2024-11-25
[Bug]: Reverse cards are due, but are not displayed for review


le-hein:
### Describe the bug

I have created quite a few flashcards on Obsidian (different notes, all properly hashtaged, with dozens of flashcards). 

I did see all of them the "Review Flashcards" pop-up window as new, I reviewed them once, a line of code generated by algorithm appeared in the notes, but by the time they became due, they never appeared in the "Review Flashcards" window again (even though the status bar mentions that there are cards due).

The problem seem to occur only to REVERSE flashcards and Cloze cards with multiple clozes {{1;; ... }} and {{2;; ... }} on one flashcard,for example. I also use Busy Siblings option; these "siblings" never appeared in the "Review Flashcards" either, even as new (after the first card was "learnt")

Am I missing something?

### To Reproduce

1. Create a single- or multiple-line reserve card/cloze card (with multiple options). -> Created a single-line reverse card yesterday, learnt it yesterday, marked it as "hard", is due today (25th November).
![image](https://github.com/user-attachments/assets/0ce1493b-cf25-4423-95fc-39bd7bce2c4b)

1.1 the same is true for cloze up cards with multiple clozes. This card should be due today, but again - it does not appear under the Review Flashcards panel.
![image](https://github.com/user-attachments/assets/ecc48f86-4cb1-4ee6-8df3-514a48321f29)

2. Open the "Review Flashcards" pop-up window on the next day. But there are no due cards for today. 
![image](https://github.com/user-attachments/assets/c65762a2-3788-40c5-badb-ec6c3220bf2a)

I tried to manually reset the card (by deleting the algorithm line), but even in this case the card doesn't appear an new anymore.

My flashcard settings:
![image](https://github.com/user-attachments/assets/3b57978d-6439-4fb3-a4aa-393b65866b28)

### Expected behavior

1. The cards that are due should appear in the "Review Flashcards" pop-up window for review.
2. Sibling cards should appear for studying/reviewing if there the only "sibling" is not new/due today.

### OS of your device

macOS Sequoia 15.1.1

### Obsidian version

v.1.7.7

### Plugin version

v1.13.2

### Installer version

v1.4.16

domtriola:
I am experiencing the same problem. I haven't found a clear pattern to identify when exactly it happens (other than only noticing it for _some_ reverse cards), so I'll share how it showed up for me in case the extra data point helps:

```md
# Words

#flashcards/debug

[...]

Issue #1192, opened 2024-12-25
[Bug]: Inline reversed flashcards in a list with custom separator not showing for review when due


tarakanbg:
### Describe the bug

I'm using a list of inline reversed flashcards with a custom separator (`=`) like this (for learning foreign words):

```
- καφενείο, το = кафене (традиционно, мъжко)
<!--SR:!2000-01-01,1,250!2024-12-06,3,267-->
- πρακτική, η = практика
<!--SR:!2024-12-23,2,243!2000-01-01,1,250-->
```

Recently, probably since last couple of updates, these no longer show as due for review, despite the comment indicating that they're past their review dates. Having the review comment on the same line does not change the behavior. However removing the list does, i.e. when not in a list, they're picked correctly for review. Not sure if the custom separator makes any difference.

### To Reproduce

- [This step might be unnecessary, possibly the behaviour is the same with the default separator]. Define a custom separator for Inline reversed flashcards in the settings (`=`). 
- Create a list with inline reversed flashcards like this:

```
- καφενείο, το = кафене (традиционно, мъжко)
<!--SR:!2000-01-01,1,250!2024-12-06,3,267-->
- πρακτική, η = практика
<!--SR:!2024-12-23,2,243!2000-01-01,1,250-->
```
- Go to review dialog and observe that no lines are due for review, despite comments indicating that both are.

### Expected behavior

 Inline reversed flashcards in a list should exhibit the same behaviour as the ones that are not in a list, or otherwise the list incompatibility should be documented.

### OS of your device

Linux

### Obsidian version

1.7.7.

### Plugin version

1.13.2

### Config file

[data.json](https://github.com/user-attachments/files/18244499/data.json)

Alebron23:
Any progress on this? I'm having a similar issue [#1190](https://github.com/st3v3nmw/obsidian-spaced-repetition/discussions/1190)

Alebron23:
@ronzulu @st3v3nmw Any thoughts?

tarakanbg:
For what it's worth: disabling the "bury sibling cards" option and the "load balancer" option in the settings **AND** manually deleting the entire `bury list` from the `data.json` file appears to have fixed the issue, although I still need to verify that after yet another flashcards review the scheduling will be picked up correctly.

PR #1199, opened 2025-01-06
fix(scheduler): clear buried questions on new day

mparq:
the logic here seemed to be half-there but maybe missed in a refactor?
the method clearIfNewDay was already implemented but not used anywhere in the codebase. the method is now called and
questionPostponementList.write now sets `buryDate` to the current date when called.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -91,6 +91,7 @@ export default class SRPlugin extends Plugin {
             this.data.settings,
             this.data.buryList,
         );
+        await questionPostponementList.clearIfNewDay(this.data)
 
         const osrNoteLinkInfoFinder: ObsidianVaultNoteLinkInfoFinder =
             new ObsidianVaultNoteLinkInfoFinder(this.app.metadataCache);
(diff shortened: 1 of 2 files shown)


mparq:
I'm not sure, but maybe fixes #1175 and #1192 (possibly others). @st3v3nmw let me know if this looks right. I am new to the codebase and it is pretty complex but the change is hopefully isolated. I just tried to solve this because I had the issue where my flashcards would never show up for review. This fix seems to fix the problem I was seeing but the nature of the fix seems to suggest that this would be broken for everybody using the bury sibling cards option?

st3v3nmw (approved):
The fix overall makes sense but I'll update the function so that we can modify `this.plugin.data` directly instead of passing it.

P.S. I think we don't need the extra change in `write` since `clearIfNewDay` will already write `data.buryDate`.

mparq:
Thanks @st3v3nmw! That makes sense. I actually don't remember why, but I was seeing that the `buryDate` was never changing from `null` without updating `this.plugin.data` in `write`... but now that I'm looking at the code again I can't figure out why that would be. Feel free to remove it, it could have just been something with how I was testing the change.

PR #1211, opened 2025-01-15
fix: Change Decks list card count box size

4Source:
Changed the minimum width of the boxes, that there will fit for numbers up to 9999. Currently there only will fit if the numbers are below 999 what in my case and certainly for others too is not enough.

Old:
image

New: 
image


st3v3nmw (approved):
LGTM, thank you!

PR #1213, opened 2025-01-18
fix: remove comments that break flashcards in the deck

Lackoftactics:
I found out that using comments sometimes breaks if flashcards are seen with patterns like quetion::answer and cloze deletion ==highlight==

This is a quick fix


Lackoftactics:
closing that one, it's easier to use callouts and don't interfere with saving data in comments by SR

PR #1233, opened 2025-02-22
feat: Apply "Option for viewing sr decks and cards in tabs" feature to Ribbon callback

bcdax110:
I wanted to try the feature in #1169, so I built it from the latest version and found that it didn’t work. Later, I discovered that it worked when  opened via command, but it didn’t work when opened via Ribbon. I guessed that the Ribbon callback might have been forgotten to be modified. So, I referred to the command callback implementation and modified the Ribbon callback part accordingly.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -538,11 +538,18 @@ export default class SRPlugin extends Plugin {
             this.ribbonIcon = this.addRibbonIcon("SpacedRepIcon", t("REVIEW_CARDS"), async () => {
                 if (!this.osrAppCore.syncLock) {
                     await this.sync();
-                    this.openFlashcardModal(
-                        this.osrAppCore.reviewableDeckTree,
-                        this.osrAppCore.remainingDeckTree,
-                        FlashcardReviewMode.Review,
-                    );
+                    if (this.data.settings.openViewInNewTab) {
+                        this.tabViewManager.openSRTabView(
+                            this.osrAppCore,
+                            FlashcardReviewMode.Review,
+                        );
+                    } else {
+                        this.openFlashcardModal(
+                            this.osrAppCore.reviewableDeckTree,
+                            this.osrAppCore.remainingDeckTree,
+                            FlashcardReviewMode.Review,
+                        );
+                    }
                 }
             });
         }


BlitzVI:
i can confirm it still doesn't seem to work if access the flashcards from the ribbon but works i use the command to access the flashcards

bcdax110:
> i can confirm it still doesn't seem to work if i access the flashcards from the ribbon but works if i use the command to access the flashcards

I have been using this modified version for a while (on Windows and iOS), and it can open as a tab from the ribbon.

Which version of the plugin are you using? If it's the releases version, it indeed won't work because the code hasn't been merged yet.

BlitzVI:
> > i can confirm it still doesn't seem to work if i access the flashcards from the ribbon but works if i use the command to access the flashcards
> 
> I have been using this modified version for a while (on Windows and iOS), and it can open as a tab from the ribbon.
> 
> Which version of the plugin are you using? If it's the releases version, it indeed won't work because the code hasn't been merged yet.

yeah im still using the release version. How can i use this modified version that u are using?

bcdax110:
> > > i can confirm it still doesn't seem to work if i access the flashcards from the ribbon but works if i use the command to access the flashcards
> > 
> > 
> > I have been using this modified version for a while (on Windows and iOS), and it can open as a tab from the ribbon.
> > Which version of the plugin are you using? If it's the releases version, it indeed won't work because the code hasn't been merged yet.
> 
> yeah im still using the release version. How can i use this modified version that u are using?

Sorry for the late reply. 

I built a version with the modifications after merging the updates from the `st3v3nmw:master` branch.

If you'd like to use it directly, you can use the main.js inside the [main.zip](https://github.com/user-attachments/files/19943777/main.zip) to replace `.obsidian/plugins/obsidian-spaced-repetition/main.js`, and re-enable the plugin in Obsidian.

If you'd prefer to compile it yourself:
1. install Node.js and pnpm(run the command`npm install -g pnpm` after install node.js ). 
2. download the code from https://github.com/bcdax110/obsidian-spaced-repetition/tree/master, and in the project directory, run the commands `pnpm install` and `pnpm build`. 
3. After that, you'll find the generated build/main.js.

Issue #1253, opened 2025-04-07
[Bug]: flashcards show up blank when context is disabled


BlitzVI:
### Describe the bug

the entire flashcard was invisible when i turned off context for the flashcards but in the settings context is referred to the only the title and headings and such

### To Reproduce

1. go settings
2.  and then go to the user interface tab
3. in the flashcards heading turn off show context in cards

### Expected behavior

it should've shown the flashcard just without the title > subheadings and such but it just made the whole card invisible

### Screenshots

With context enabled
![Image](https://github.com/user-attachments/assets/953911b7-9fc3-41dc-a69e-c9ca6ce67542)

without context enabled
![Image](https://github.com/user-attachments/assets/eae6a615-9c5c-405b-87ce-37bf4dd0a0af)
its just blank

### OS of your device

win11

### Obsidian version

latest

### Plugin version

latest

### Installer version

v.1.7.7

bl0ggi:
Same issue.

mtomas7:
To fix it manually, implement #1281 

Find the file under: `\Your-Note-Folder\.obsidian\plugins\obsidian-spaced-repetition\main.js`

It will be line 27708 starting with: `_updateCardContext() {`

After this line insert new line with code: `if (!this.cardContext) return;`

So the whole block will look like:

```
  _updateCardContext() {
    if (!this.cardContext) return;
    if (!this.settings.showContextInCards) {
      this.cardContext.setText("");
      return;
    }
    this.cardContext.setText(
      ` ${this._formatQuestionContextText(this._currentQuestion.questionContext)}`
    );
  }
```

cspipaon:
@mtomas7 If implementing the fix, should we be aware of anything when the plugin next updates? 

For example:
- could it cause any issues with the update? 
- could the update eliminate the snippet?

st3v3nmw:
Hi! This should be fixed in the latest version 1.13.4.

PR #1281, opened 2025-06-19
Add check to make sure card context exists

### LINKED ISSUE #1250: [Bug]:  not show any cards
### Describe the bug

after the update does not show any cards, the window is empty

### To Reproduce

open plugin 
click on flashcards
we see an empty space instead of 32 cards

### Expected behavior

it's obvious we need to see the cards

### Screenshots

_No response_

### OS of your device

win11

### Obsidian version

1.8.9

### Plugin version

1.13.3

### Installer version

_No response_

### Additional context

_No response_

### Config file

_No response_

bgreenawald:
Checks that the `cardContext` exists before trying to set it. Resolves the issue where disabling the **Show context in cards** setting would cause the flashcard not to load. 

Closes #1250 and #1253

### DIFF
--- a/src/gui/card-ui.tsx
+++ b/src/gui/card-ui.tsx
@@ -440,6 +440,7 @@ export class CardUI {
     }
 
     private _updateCardContext() {
+        if (!this.cardContext) return;
         if (!this.settings.showContextInCards) {
             this.cardContext.setText("");
             return;


st3v3nmw (approved):
Thanks for the fix!

PR #1283, opened 2025-06-28
[FEAT] Add Support Callout Format Flashcards

currently-coding:
This PR adds support for callout formatted flashcards. 
## Why

Imagine you're preparing for an exam and want to review all your flashcards for a specific topic, even if they're not currently due. You open the plain `.md` file where they’re stored, but:

- With the standard format (`A :: B`), the answer is immediately visible — there's no way to hide it. This undermines the recall-testing effect, as your eyes catch the answer before you can think.
- Additionally, the `A :: B` format isn’t ideal for manual revision — it's visually dense and harder to process.

By using Obsidian’s native **callouts**, flashcards can be stored in a **folded state**. This allows you to:
- View only the question initially.
- Manually unfold the answer after attempting recall.
- Maintain a clean and structured layout for studying.
- Still use the features of regular spaced-repetition.

## Example

```obsidian
>[!Question]- What is 2+2?
>4
```
image
can now be interpreted as a normal, one-way flashcard.
image

## How is this implemented?

- introduced a new Flashcard type `Callout` and respective Settings `calloutCardMarker` and `calloutLineMarker`
- updated parse() funktion to detect callouts
- added CalloutType() to QuestionTypeFactory()
  - Removes the callout syntax parts(e.g. ">" from the start of a line and "[!Question]+")
  
Note: I did not change anything regarding UI, so currently only the defaults I set work.

## Testing

- Added unit tests inspired by the ones for basic multiline flashcards in:
  - `parser.test.ts`
  - `question-type.test.ts`
 
## Examples

image
image
image

--- 
(this is my first ever pull-request, just ask if you need anything I've missed to list)


currently-coding:
Perhaps it should support different callout-types besides `>[!Question]` so that the user can decide what "callout-tags"(I don't know what they're called) will be interpreted as a flashcard. Personally I use `>[!Question]` and `>[!Definition]`. I might add that to the PR in the next days.

PR #1306, opened 2025-08-28
fix: prevent error when showContextInCards is disabled by checking cardContext existence

CodeXiaoMai:
If I turn off the "show context in cards" option, there will be nothing left on the cards, not even the question.


st3v3nmw:
Hi,
Thanks for the PR. I've merged #1281 instead which was opened earlier than this one.

PR #1338, opened 2025-12-13
feat: Adds a button to jump to the current item in your notes

### LINKED ISSUE #1228: [FEAT]: please make move button.

The leftmost button currently opens a card edit modal when clicked. 
Editing cards in the modal window is inconvenient.

Image

Instead of this behavior, I would like to change it so that clicking the button moves the cursor to the location/position of this card in the note.

### Additional context

_No response_

sheldoncork:
Adds a new button to jump to where the card is in your notes for easy modification.

https://github.com/user-attachments/assets/8a50d7c7-9a0c-4e49-93fa-1789232ade36

An alternative solution would be a setting that overrides the editButton to jump to the note instead.

Closes #1228

### DIFF
--- a/src/gui/sr-modal.tsx
+++ b/src/gui/sr-modal.tsx
@@ -1,4 +1,4 @@
-import { App, Modal, setIcon } from "obsidian";
+import { App, MarkdownView, Modal, setIcon } from "obsidian";
 
 import { Deck } from "src/deck";
 import {
@@ -79,6 +79,7 @@ export class FlashcardModal extends Modal {
             this.contentEl.createDiv(),
             this._showDecksList.bind(this),
             this._doEditQuestionText.bind(this),
+            this._jumpToCurrentCard.bind(this),
         );
     }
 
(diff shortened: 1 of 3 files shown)


src/gui/sr-modal.tsx
```diff
@@ -140,6 +141,31 @@ export class FlashcardModal extends Modal {
...
 
+    private async _jumpToCurrentCard(): Promise<void> {
+        const currentQuestion = this.reviewSequencer.currentQuestion;
+        if (!currentQuestion) return;
+
+        const file = currentQuestion.note.file.tfile;
+        const blockId = currentQuestion.questionText.obsidianBlockId;
+        const line = Math.max(0, currentQuestion.lineNo ?? 0);
+
+        this.close();
```
KyleKlus: Hi, so I think it would be great, if it either doesn't apply to the modal view or if we dont close the modal here so, that it just opens the note in the back ground and doesnt stop the whole review process

PR #1385, opened 2026-02-26
feat: add gamification support with Habitica integration

### LINKED ISSUE #1384: [FEAT]: Gamification

It would be dope if Spaced Repetition supported gamification.

My proposed approach would be to integrate an external gamification tool, such as Habitica. Reviewing the card would trigger scoring a dedicated task/habit in the external tool, rewarding the user for engaging in spaced repetition and active recall.

### Additional context

Similar request #1193

byq77:
Closes #1384

Description: TODO

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,90 +2,127 @@
 
 <img src="https://img.shields.io/github/downloads/st3v3nmw/obsidian-spaced-repetition/total" /> <img src="https://img.shields.io/github/downloads/st3v3nmw/obsidian-spaced-repetition/latest/total?style=flat-square" /> <img src="https://img.shields.io/github/manifest-json/v/st3v3nmw/obsidian-spaced-repetition?style=flat-square" /> <img alt="Codecov" src="https://img.shields.io/codecov/c/gh/st3v3nmw/obsidian-spaced-repetition">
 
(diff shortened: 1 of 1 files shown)


st3v3nmw:
Thanks for the proposal and the effort you put into this.

I'm wary of merging it though, since it adds a hard dependency on a third-party API, which means maintenance burden every time Habitica changes something on their end.

The underlying motivation makes sense to me, but I'd rather address it natively, think streak tracking and a review heatmap along the lines of what Anki does. That keeps everything self-contained.

I did briefly consider whether a plugin extension system could be the right answer here, but given that this is already an Obsidian plugin, I'm not sure that's a rabbit hole worth going down.

byq77:
> Hi, this pr will be closed in 3 months from now (on the 23rd of june 2026), if there is no change in plans of adding this by then. Sry for your lost effort, but we have to keep this plugin maintainable 😇

No worries, that's what forks are for. Personally, I am a long-term Habitica user, and I enjoy gamifying different experiences for personal use. However, Spaced Repetition would benefit greatly from a gamification add-on, as it improves user engagement. Imagine the Xbox 360 experience without achievements. It would suck. But that's just my opinion.

byq77:
Closing, because maintainers have a different plan for tackling the gamification feature.

PR #1432, opened 2026-03-28
fix: close modal on jump

sheldoncork:
Closes modal when jumping to a card.
Adds translation for card.

Also fixes "undefined" was string instead of actual undefined.


KyleKlus:
Hi, are you still working on this😊?

sheldoncork:
I have not touched it since. Seems like for the modal it would be two buttons then and the tabbed version left as is. I'll make changes

KyleKlus:
Yep, I think that is about right; thank you for continuing this 😊

sheldoncork:
Should be good to go. I see the commits in history follow the semantic style deal. Should I rebase these to fix them to follow that format?

KyleKlus:
Thank you for implementing this 😊.
No need to adjust the commit style; only the PR style is relevant 😇

PR #1458, opened 2026-04-10
fix: improve keyboard navigation for cloze cards

EdibleSalt:
Problems
When reviewing cloze cards, two keyboard interaction issues made the workflow inefficient:

1. After typing in the cloze input, pressing Enter did nothing so users had to use Tab+Space or the mouse to confirm.
2. After rating a card, focus was lost, so users had to click back into the review window before keyboard shortcuts worked again. Additionally, the cloze input was not automatically focused for the next card.

Changes
In card-container.tsx:

1. Added a keydown listener on each .cloze-input element that calls _showAnswer() on Enter
2. Added auto-focus on the first .cloze-input at the end of _drawContent() for a fully keyboard-driven review workflow

### DIFF
--- a/src/ui/obsidian-ui-components/content-container/card-container/card-container.tsx
+++ b/src/ui/obsidian-ui-components/content-container/card-container/card-container.tsx
@@ -233,6 +233,13 @@ export class CardContainer {
 
         // Setup cloze input listeners
         this._setupClozeInputListeners();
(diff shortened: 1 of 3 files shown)


KyleKlus:
Thank you for your contribution 😊. This really is less cumbersome with your improvement 😇

PR #1462, opened 2026-04-15
fix: fix "Jump to cart" feature

byq77:
This PR fixes an issue with the "Jump to card" feature by preventing duplicate tabs from opening for the same file. Now, if a file is already opened in another tab (leaf), it will be focused and scrolled to the correct line instead of opening a new tab.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,8 +5,14 @@
 # vscode
 .vscode
 
+# devcontainer
+.devcontainer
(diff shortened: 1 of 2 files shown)


KyleKlus:
Thank you for your contribution 😊. 
This was quite the oversight of me to not check if what happens if the note is already open 😅

Issue #1465, opened 2026-04-19
[FEAT]: Optional sidecar storage for schedule data


MaxWolf-01:
### Is your feature request related to a problem? Please describe.

Schedules are currently stored as `<!--SR:!date,int,ease-->` comments inline with each card. This mixes content with the scheduling state of that content, and the cost shows up in lots of small ways:

- Every review mutates the note (noisy git diffs, friction when editing).
- Shared/published notes carry schedule data they shouldn't.
- Notes fed to LLMs or embeddings carry noise.
- Trying another SRS plugin or algorithm against the same notes is harder than it should be.

None of that is *required* by the design — the `IDataStore` interface and the `DataStoreName` enum (currently only `NOTES`) already anticipate alternatives, and the roadmap in #1397 mentions "different storing options" as a goal.

### Describe the solution you'd like

A `StoreInSidecar` implementation of `IDataStore`, opt-in via settings:

- One sidecar JSON per note, mirror-tree under `.obsidian/plugins/obsidian-spaced-repetition/schedules/<path>.sr.json`. Keeps the vault clean; syncs with `.obsidian/` like any other plugin data.
- Schedules keyed by the existing `textHash` of the card's front text (already computed in `question.ts` for the bury list). Works for every card type the plugin supports. Editing the front invalidates that card's schedule — acceptable cost, explicit user action to reset.
- No note mutation of any kind — no block IDs, no frontmatter.
- Note rename/delete handled via `vault.on('rename' | 'delete')`.
- Orphaned schedules (front edited, card removed) surfaced via a cleanup view.

KyleKlus:
Hi, I would love to implement this in the future 😊. 

Right now I am heavily focusing on dev experience (Sometimes also ui stuff as self care when working on this project 🤭), as a lot of bad/unmaintainable code has accumulated over the years😅. So it might take a while till I get around to this 😇

PR #1471, opened 2026-04-23
feat: store scheduling data in plugin data instead of inline in notes

### LINKED ISSUE #162: [FEAT] Stop using YAML Move plugin info and data to separate file

Lorite:
## Summary

- Adds an optional **Store in plugin data** mode (beta) that keeps all scheduling metadata in a single plugin data file instead of writing `<!--SR:...-->` comments and `sr-*` frontmatter into notes
- Card schedules are keyed by content hash (`cyrb53(topicPath + questionText)`), so they survive renames and moves — including those done by external tools outside Obsidian
- Note schedules continue to use the file path as key; a `vault.on('rename')` listener keeps them in sync for Obsidian-internal renames
- Switching modes triggers a confirmation modal and **fully migrates** all existing scheduling data in both directions:
  - **NOTES → PLUGIN_DATA**: reads all `<!--SR:...-->` comments and `sr-*` frontmatter, writes to the plugin data repository, then strips the metadata from files
  - **PLUGIN_DATA → NOTES**: reads all schedules from the repository and writes them back as inline comments and frontmatter, then clears the repository
- A one-time startup migration rewrites any legacy `notePath::hash` card keys to hash-only keys

## New files

| File | Purpose |
|---|---|
| `src/data-stores/plugin-data/schedule-data-repository.ts` | CRUD layer over the persisted `scheduleState` object |
| `src/data-stores/plugin-data/plugin-data.ts` | `IDataStore` implementation for plugin-data mode |
| `src/data-store-algorithm/data-store-in-plugin-data-algorithm-osr.ts` | Returns empty string so no SR comments are written to notes |
| `src/data-stores/data-store-migrator.ts` | Bidirectional migration logic between the two modes |

## Test plan

- [x] All 360 unit tests pass
- [x] `ScheduleDataRepository` — full coverage: note/card CRUD, state initialisation, legacy key migration, `renameFile`, `clearState`
- [x] `StoreInPluginData` — all public methods covered including fallback-to-text and external-data-takes-precedence paths
- [x] Manual: enable "Store in plugin data" in settings, confirm migration modal, review cards and notes, verify no `<!--SR:...-->` comments appear in files
- [x] Manual: switch back to "In the notes", confirm migration modal, verify `<!--SR:...-->` comments reappear
- [x] Manual: rename a note inside Obsidian and verify schedule is preserved
- [x] Manual: rename a note with an external tool and verify card schedules are preserved (note schedule will be lost — frontmatter UUID follow-up planned)

## Known limitations / follow-up

- Note review schedules are still keyed by file path; external renames will lose the note schedule. A frontmatter UUID approach is planned as a follow-up
- The mode is intentionally labelled **beta** in the UI

## Kyles changes

- [x] Refactors code into one schedule part and one data part
- [x] Generalizes decks for merge with reviewable notes
- [x] Fixed a **LOT** of miss typed variables, constants & functions 😮‍💨
- [ ] Implement to use existing sr-id even if there is no schedule for it
- [ ] Re implementing plugin data store with new refactored code
- [ ] Implement a data view key value export for the folder storage?
- [ ] Re implementing folder data store with new refactored code
- [ ] Re implementing "delete all schedule data" feature for each data store


KyleKlus:
Hi, thank you so much for implementing this😊.

3 things: 
- I think it is very important to tell the user that any rename of the note will break the scheduling data, as long as there isn't a way to bind the scheduling data to something else than the note file path
- It is very important to maybe link the scheduling data via a UUID, before we even offer this feature, as people are probably renaming/moving their notes quite often
- It would be awesome to implement more tests, to get the coverage up so that the codecov report isn't failing 😇

P.S.: Maybe have a look at the issues which I linked up, to see if they offer a good solution for the schedule identification problem

KyleKlus:
Oh and the data store option has moved to the data settings page to make it more organized 😇

Lorite:
Addressed all three points from the review:

**1 & 2 — UUID-based note identification (replaces the rename warning)**

Rather than just warning users about renames, I implemented the UUID approach. Each note now gets an sr-id field written to its frontmatter the first time it is scheduled in plugin-data mode (crypto.randomUUID()). Because the UUID travels with the file, note-level scheduling data now survives any rename or move — including ones done outside Obsidian entirely. The info text in the settings explains this to the user, and the migration confirmation mentions that sr-id will be added to frontmatter.

Backward-compat: existing path-keyed entries (from before this commit) are lazily migrated to UUID on the first read. The vault rename listener is kept for any path-keyed entries that haven't been read yet.

Card schedules remain keyed by content hash — this limitation (editing a card's front text resets its schedule) is now surfaced in the settings info text.

**3 — Test coverage**

- Added DataStoreInPluginDataAlgorithmOsr test (was 0% coverage).
- Added 3 tests to ScheduleDataRepository covering UUID-keyed storage and the interaction between UUID keys and the rename handler.
- All 364 tests pass.

**Data settings page**

Moved the data storage dropdown and all delete-scheduling-data buttons into a new dedicated "Data" settings page (data-page.tsx), as suggested. The scheduling page now only contains algorithm/scoring settings and the inline-comments toggle.

Lorite:
I rebased and force pushed :)

KyleKlus:
Hi, so the code coverage still says that some lines aren't covered by tests, so it would be nice if you had another look at those😇

Here is the most recent codecov report:

https://app.codecov.io/gh/st3v3nmw/obsidian-spaced-repetition/pull/1471

KyleKlus:
Is the data saved in the data.json of the plugin or in a separate file?

I think it would be great if it is in a separate file, because the data.json is more right now like a config file😇.
Maybe have a look if a approach like mentioned in #1465 is possible🤔.

KyleKlus:
Hi, so I had a look at the possibility of storing the schedule information outside of the data.json.
I found out, that we can't store any other file in the plugin folder inside of the .obsidian folder for compatibility reasons with obsidians sync service 😅.

So maybe we could store the scheduling data in a folder inside the vault, like excalidraw does it for their user defined scripts.
This way we have a user defined location (by default "/Spaced Repetion") and there we have "Schedule Data" folder, where we store the data in json format, but in a markdown file, as json files aren't synchronised by default with the obsidian sync service. So the file would maybe be called "card-schedule-data.sr.md" & "note-schedule-data.sr.md".

What do you think of this 😇?
If it is all too much to implement for you with the extra file location, then it would also be fine for me to just store it in the data.json and then I'll implement that with the extra file location later on 😊.

Lorite:
I'll try to find some time tomorrow or this week to implement the separate file solution and come back to you with the 100% test coverage as well

KyleKlus:
Alright, but you dont need to rush with it, as I'll also be quite busy this week and wont really have time to review it 😇

Sejder:
> So maybe we could store the scheduling data in a folder inside the vault, like excalidraw does it for their user defined scripts. This way we have a user defined location (by default "/Spaced Repetion") and there we have "Schedule Data" folder, where we store the data in json format, but in a markdown file, as json files aren't synchronised by default with the obsidian sync service. So the file would maybe be called "card-schedule-data.sr.md" & "note-schedule-data.sr.md".

as a nextcloud and git user for my obsidian syncs it would be very nice to have the option to choose the data location such that it can work with obsidian sync inside the obsidian project itself, but also with the other sync options such that it does not clutter the project and is hidden nicely in .obsidian. No rush on this but i think it would be a fairly simple feature with a lot of value:)

KyleKlus:
> > So maybe we could store the scheduling data in a folder inside the vault, like excalidraw does it for their user defined scripts. This way we have a user defined location (by default "/Spaced Repetion") and there we have "Schedule Data" folder, where we store the data in json format, but in a markdown file, as json files aren't synchronised by default with the obsidian sync service. So the file would maybe be called "card-schedule-data.sr.md" & "note-schedule-data.sr.md".
> 
> as a nextcloud and git user for my obsidian syncs it would be very nice to have the option to choose the data location such that it can work with obsidian sync inside the obsidian project itself, but also with the other sync options such that it does not clutter the project and is hidden nicely in .obsidian. No rush on this but i think it would be a fairly simple feature with a lot of value:)



> > So maybe we could store the scheduling data in a folder inside the vault, like excalidraw does it for their user defined scripts. This way we have a user defined location (by default "/Spaced Repetion") and there we have "Schedule Data" folder, where we store the data in json format, but in a markdown file, as json files aren't synchronised by default with the obsidian sync service. So the file would maybe be called "card-schedule-data.sr.md" & "note-schedule-data.sr.md".
> 
> as a nextcloud and git user for my obsidian syncs it would be very nice to have the option to choose the data location such that it can work with obsidian sync inside the obsidian project itself, but also with the other sync options such that it does not clutter the project and is hidden nicely in .obsidian. No rush on this but i think it would be a fairly simple feature with a lot of value:)

Hi, so storing it within the .obsidian folder won't be possible, as it will not be synced by the obsidian sync service. Another limitation is that on mobile the access to files will only be reasonably possible via the obsidian api, which means we can only create new files in the actual vault.


So I think a good compromise would be that we store it in a customizable location (by default in "/Spaced Repetition") within the actual vault to keep everything compatible 😇. This folder should also only be created, if the user selects the option to store the schedule in a separate file.

KyleKlus:
Hi, so I had a bit of a look around and I noticed the following:

- The folder got created right away before I even turned on the setting to store it in the folder -> I think it would be better to only create it after the user decided to use the new data storage option
- Then right after the plugin created the folder (it wasn't there before) it fully crashed the plugin, because it had an error which said that the folder already existed 😅
- I think it is also a bit problematic that the schedule is lost, when the card changes, as one could have a simple spelling mistake to correct and then the whole progress is lost 
    -> Ideally I think we should find a better way to identify the card 🤔
    -> But if we really choose to go down that route that the schedule is lost on edits within the note, then I think it would be at least good to offer an edit functionality through the edit modal and then reconnect the schedule to the updated card. This would be possible there, as we know which card we are editing through the modal and thus its schedule 😇

So for now I would pick the latter option as a tradeoff for a less cluttered note, but I would definitely offer the edit modal solution and highlight it very clearly that this trade off was made.

Lorite:
Hi @KyleKlus @MaxWolf-01, I won't have time to work on it for a while... But feel free to edit my PR yourselves if you find the time! :)

KyleKlus:
> Hi @KyleKlus @MaxWolf-01, I won't have time to work on it for a while... But feel free to edit my PR yourselves if you find the time! :)

No worries, I am also not in a rush 😇.
I'll also have to see when I find the time to work on this😅; but thank you for implementing it this far😊.

MaxWolf-01:
if losing the schedule on card edit is fine, would note rename need a uuid? path + the rename listener already handle in-obsidian renames, external could reset like a card edit. then sr-id might not be needed / we can avoid yaml

KyleKlus:
The uuid is needed because people could also rename files out side of obsidian running and then we would loose the connection between card and schedule 😇.

I could make the uuid optional for people to whom it is more important to have no yaml than to have consistency between external renames🤔.

MaxWolf-01:
I would appreciate that, because I don't generally rename outside of Obsidian, and if I do, the schedule could be quickly restored.

KyleKlus:
Alright, then I think we have found a common ground for all the requirements of this feature😊:

##### How to migrate the files once the user decides to use this feature:
- The required folders will be created once the user changes the option
- The schedule data will be migrated all at once with plenty of warnings and progress visuals

##### How to migrate the files once the user decides to not use this feature:
- The folders will stay till the user deletes them
- The schedule data will be migrated all at once with plenty of warnings and progress visuals

##### How & where to store the files:
- In the vault in a customizable location (by default in "/Spaced Repetition"), as the obsidian sync service doesn't sync random files in the plugin folder by default and because it is hard to create/update files in the plugin folder on mobile
- In that folder is a non customizable folder named "Schedule Data"
- The card schedule data will be stored in a file named "card-schedule-data.sr.md" which is located in the "Schedule Data" folder
- The note schedule data will be stored in a file named "note-schedule-data.sr.md" which is located in the "Schedule Data" folder
- The data is stored in json format, but still in a markdown file, so that the obsidian sync service syncs it by default
- The data is linked to a note/card either via the path of the note or via a uuid (which is more robust against renames)

| Pro                            | Con                                                        |
| ------------------------------ | ---------------------------------------------------------- |
| Obsidian sync works by default | Storing the files in the .obsidian folder wont be possible |
| Obsidian sync works by default | Storing the files in .json file wont be possible           |



##### How to link cards to the scheduling data with a uuid:
- A uuid will be set in the frontmatter for each note with cards in it, in order to link the data to a note 
- The already existing hash of a question is used to link the data to the card

| Pro                                                         | Con                                                                                                                  |
| ----------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------- |
| The link is more robust against renames and moving the note | uncontrolled deletion of the uuid or the note breaks the link, uncontrolled modification of the uuid breaks the link |
| there wont be scheduling comments                           | uncontrolled modifications of the card breaks the link                                                               |


##### How to link cards to the scheduling data without a uuid:
- The path of a note will be used to link the data to a note 
- The already existing hash of a question is used to link the data to the card 

| Pro                                | Con                                                             |
| ---------------------------------- | --------------------------------------------------------------- |
| There wont be any frontmatter data | uncontrolled deletion/moving/rename of the note breaks the link |
| there wont be scheduling comments  | uncontrolled modifications of the card breaks the link          |

##### How to preserve the link between a card/note and its scheduling data, with uuid:
- Edits of the card via editor or the edit modal have to be monitored to actively keep the link alive

##### How to preserve the link between a card/note and its scheduling data, without uuid:
- Modifications of the note have to be monitored to keep the link alive 
- Edits of the card via editor or the edit modal have to be monitored to actively keep the link alive



It will be quite tedious and time consuming to implement all of this 😅

st3v3nmw:
@KyleKlus, I might be missing something but isn't using the question hash fragile too? If someone changes their question (even slightly), all of a sudden the link is gone. One idea I had a while back was generating block ids, but something shorter like https://www.npmjs.com/package/short-uuid (instead of a full uuid), and then using the Obsidian ^blockid notation. Say with a cloze deletion:

```
asdf ==asdf== asdfwe asdf apomine ^blockid
```

The cloze is simple enough, other question types might be complicated.

KyleKlus:
@st3v3nmw -> Yes, that is one of the trade offs we have to make here 😇. My plan is to warn the user about this trade off several times and also catch any edits of the card as best as I can (Sort of a best effort service) to minimize card breakage.
The store data in file option is mainly for people who are annoyed with having any sort of scheduling comment in their notes. So I think the option with using block ids will also not be good enough 😅.

KyleKlus:
Later on (in another pr) I plan to add an option to use ````$\phantom{<--SR:...-->}$```` as a formatting for the sr comment. This is a latex way of turning text invisible😇

KyleKlus:
Maybe I could also add a way to collect orphaned cards which have been disconnected from their note, so that the user can reconnect the schedule manually somehow🤔

KyleKlus:
> Later on (in another pr) I plan to add an option to use `$\phantom{<--SR:...-->}$` as a formatting for the sr comment. This is a latex way of turning text invisible😇

Although I might go the route of creating a custom callout just for the purpose of hiding the sr schedule 😇, as it seems less hacky and more legit looking with the edit code button on the right side🤔

st3v3nmw:
Hmm, okay 🤔. If people don't want any scheduling metadata whatsoever, then we'd have to do that. Though I still think the question hash approach has lots of footguns that'll probably lead to confusion/frustration to someone using the plugin. Like say, they want to make changes to their cards outside of Obsidian. Or if the OSR listener watching file changes fails for some reason, they lose their scheduling.

A card id like `<!-- id: a3f7 -->` or `^blockid` is a one-time write that makes the system reliable. It's ugly but the ID lives with the content, moves if the file moves, and the user can see it. That's a better trade than a fragile system with defensive messaging around it. And the listener approach isn't free either: if the underlying Obsidian API changes, suddenly a lot of things break. We probably don't need the collision-resistance guarantees of a full UUID either.. Obsidian's own block references are 6 alphanumeric chars, which gives over 2 billion combinations which is plenty of headroom.

The linked issue #162 mentions some use cases (sharing notes with others, reducing the blast radius of changed files after a review, etc), and these can be done with a card/block id. While I'd also love to not have any scheduling metadata whatsoever in the notes, something like this is probably the more predictable model both for us and OSR users.

Otherwise I fear we'll be fielding issues down the line when someone inadvertently loses their scheduling, and "we warned you" is cold comfort when months or years of review history disappears because they fixed a typo 😅.

KyleKlus:
Yeah I totally get you with the fragile system and its complexity for both users and developers🤔. 
Thank you for considering/highlighting all the issues 😇.

Alright, then I'll first focus on separating the schedule from the card first and then later on in a different pr I'll maybe have another go at the id visibility issue. So the approach I am going for now will be:

- Each reviewable card gets a block id like this ("sr-id" is used to make them easy to find)-> ^sr-id-[generated short uuid goes here]
- Each reviewable note gets a sr-id in the frontmatter like it is already sort of done 
- Those ids link then to the schedule data files

KyleKlus:
@Lorite Sry for pirating your pr for a refactoring of a large part of the backend code😅. 

I just find it very difficult to work with the current state of the general backend code, as the files are not really in a hierarchical folder structure, most of the types are wrong/sketchy and not really well documented🤭

Lorite:
No worries at all @KyleKlus 😁 it's great that you guys took over the PR. Feel free to change it as you please

Merkwurdichliebe:
As I've grown to depend on this plugin in the last month, and this feature is the one that I'd like most to see implemented, I'd like to add something that happens on my end quite often.

Sometimes I need to reorder a note's flashcards, move some text around etc, and it's handy to have the sr comment just next to it, so that progress isn't lost after editing.

I would very much like to still be able to modify a card (sometimes in batch via VSCode) without it disrupting SR progress. Keeping the data easily editable and not burying it too deep in a settings file could be advantageous to some users. I know there isn't an ideal solution to this and moving everything to a centralised file is the way to go of course. This is just something to keep in mind, I appreciate very much what everyone here is doing already.

Thanks.

KyleKlus:
Hi, I think this should still work with the current way which I am implementing this 😊. There will still be a small block id at the bottom of the card, which links the card to its schedule.

In addition to this there will be an optional custom callout wrapped around it, which visually hides away any id/scheduling data😇

Merkwurdichliebe:
> In addition to this there will be an optional custom callout wrapped around it, which visually hides away any id/scheduling data

Guess I'll have to rewrite my custom CSS again :)

> There will still be a small block id at the bottom of the card.

So if I understand correctly, a user's .md file will only contain different card ids in a syntax similar to the current one, but these will not be changing over time? If so, it sounds perfect. Just my 2c.

Also, I don't know if it's planned (or done already), but please allow *deferring* the conversion to the new format in case the user needs to learn about how it looks before committing (i.e. don't prompt for conversion after the update and then not allow it anymore). Probably obvious, but just in case.

KyleKlus:
Yeah, that is how it is planned 😇.

All migrations between the data storage options that we offer will happen via the settings. There will be plenty information on how each data storage option looks and functions. Then there will be a confirmation modal & a progress modal when the user decides to switch/migrate to one of the other data storage options 😊

KyleKlus:
You wont have to change your css as long as you don't switch to one of the external data storage options. The current format will stay how it is, as we try to keep most of the features optional 😇.

I also think it might be easier to customize the css, with the new callout data syntax, as now one has some css container to select the scheduling data part 😊

Merkwurdichliebe:
Oh I don't mind modifying the CSS, it's an ongoing project :) I'm looking forward to the new data storage option - like many others I find that modifying the file's timestamp regularly is annoying.

It's not much currently, if you want to see a real-life use case.

```
/* In notes that use flashcards,
hide the <br> following the %%card%% separator comment,
so as to eliminate the single-line vertical separation
between card front and back when viewed as a regular note.*/
.flashcards .el-p:has(strong) br:first-of-type {
    display: none;
}
```

I also need to hack something like this because some of my notes have JS-generated info which I need to manually convert to a (hidden) flashcards using Templater when I first create them.

```
/* Hide the flashcard-only section at the bottom of the note */
.study .node-insert-event + .el-p:has(mark) {
    display: none;
}
```

Hopefully all of this will be easier to do once flashcards get a container.

Thanks again for your work.

KyleKlus:
Hi, I have some bad news 😮‍💨.

I tried my hardest to implement this, but I cant get it to work with the current state of the code base. There are just too many random undeclared null values to make this code properly debuggable. I have tried for the whole past 2 weeks to make sense of this, but I have to cut my losses here.

So I am going to merge this as a PR which refactors the code base and which fixes a lot of miss typed code, as this PR fixes a lot of those bugs/errors. The data store feature has to wait till someone else picks up on this, as I just cant spend more energy into this bottomless pit.

I am really sry for everyone who had hopes that this will be implemented now and for everyone who worked on this.
I have packaged the half implemented data store code in a zip file and I have documented how to pick it up again in a README next to it.

I am also planning on creating a issue which documents all of this so that someone else can easily continue with this again.

PR #1244, opened 2025-03-15
feat(flashcards): Added option to exclude tags from being considered for Decks

Hugo-Persson:
Added new option to enter tags to ignore for Flashcard tag deck creation.

This is useful for ignoring old decks without needing to alter the notes. 

Solves: https://github.com/st3v3nmw/obsidian-spaced-repetition/issues/737

PR #1486, opened 2026-04-29
feat: add "tags to ignore" for flashcards and notes

### LINKED ISSUE #737: [FEAT] Exclude flashcard tags to review
I have certain flashcard decks that are "immature", i.e. not fully fleshed out.

There should be an option to disable tags, similar as there is an option to enable tags.

A bad substitute could be to use a tag-changer plugin to change tags from a "staging" tag to the official flashcard tag, but these plugins are dangerous as they edit your file contents with string replacement and can ruin your whole deck.

So, better to ignore certain decks.

### LINKED ISSUE #1121: [Bug]Folders to ignore edit box automatically exits
### Describe the bug

There is an issue when I set folders to ignore under the _Flashcards_ and _Notes_ menu.
Once I click the edit box and enter characters, it automatically exits the edit box after about 0.8 seconds.

### To Reproduce

1. Go to the Spaced Repetition settings
2. Click on the edit box of _Folders to ignore_
3. Enter any character(one or more)
4. Waiting for about 0.8s
5. See error

### Expected behavior

The edit box won't automatically exit.

### Screenshots

_No response_

### OS of your device

MacOS 14.6.1

### Obsidian version

v1.6.7

### Plugin version

v1.13.0

### Installer version

_No response_

### Additional context

_No response_

### Config file

_No response_

### LINKED ISSUE #1186: [FEAT]: Being able to archive decks

I have been using this plugin for a long time and I love it, but a problem I am facing is a growing library of cards. I would like to be able to archive certain decks to not show up when reviewing all decks. 

A setting that allows the user to input decks that should be excluded from review. 

### Additional context

I know that it is possible to exclude folders, but all my notes are in one folder. 

I would love to help to build this if it is a good feature that will be accepted

flowing-abyss:
Adds a "Tags to ignore" field in both the Flashcards and Notes settings pages, right before "Folders to ignore". Notes carrying any of the specified tags are fully skipped — no cards loaded, no notes added to the review queue.

This is useful when some notes match an include tag but shouldn't be reviewed yet — drafts, archived material, work-in-progress decks — without touching the notes themselves. Both fields use the same prefix-matching as the include-tag fields, so `#archived` also covers `#archived/old`.

Closes #737, #1121, #1186

---

I'm aware of #1244 which addresses the same issue, but the behavior is different: it excludes tags from deck creation while the cards still appear in review. This PR goes for full exclusion, which I think matches the intent of #737 more closely. It also extends the same idea to the note review queue, which #1244 doesn't cover.

The implementation follows the current v1.14.1 architecture, all existing tests pass, and the linter is happy. Unit tests for the new `SettingsUtil` helpers are included.

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -115,10 +115,12 @@ export class OsrCore {
         // (Performance note: This only requires accessing Obsidian's metadata cache and not loading the file)
         this.osrNoteGraph.processLinks(noteFile.path);
 
+        const tags = noteFile.getAllTagsFromCache();
+
         // Does the note contain any tags that are specified as flashcard tags in the settings
         // (Doing this check first saves us from loading and parsing the note if not necessary)
         const topicPath: TopicPath = this.findTopicPath(noteFile);
-        if (topicPath.hasPath) {
+        if (topicPath.hasPath && !SettingsUtil.isAnyTagIgnoredForFlashcards(this.settings, tags)) {
             note = await this.loadNote(noteFile, topicPath);
             if (note !== null) note.appendCardsToDeck(this.fullDeckTree);
         }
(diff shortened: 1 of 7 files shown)


KyleKlus:
Hi, thank you for implementing this 😊. 
I'll take a look at it once I find the time, as this week is quite a full week for me, 
so I might only get around to it next week 😇.

KyleKlus (commented):
I had a look at the code and so far, everything looks good😊. Now I just have to go over it in the ui itself😇.

flowing-abyss:
Thanks! Here are a couple of screenshots showing what it looks like in practice — both the Flashcards and Notes settings pages.

image
image

KyleKlus:
Thank you for your contribution to the plugin😊

PR #1513, opened 2026-05-13
feat: auto-open review sessions for ready flashcards

### LINKED ISSUE #1512: [FEAT]: Auto-open review sessions for ready flashcards
## Is your feature request related to a problem? Please describe.

Users can miss scheduled flashcard reviews when they do not manually open the review command during the day. This is especially easy when Obsidian stays open in the background while there are new or due cards ready to learn.

## Describe the solution you'd like

Add an opt-in setting that periodically checks for new or due flashcards and automatically opens a review session when cards are ready. The session should reuse the existing review flow and deck selection logic rather than implementing a separate review UI or scheduling algorithm.

The feature should avoid interrupting active review/editing states, open directly into a deck/card flow where possible, remember the last reviewed deck, and provide configurable attention cues such as a notice, alert sound, and dock bounce.

## Additional context

A PR is being opened with the implementation. The feature is disabled by default and includes per-cue controls so users can decide how intrusive the auto-open behavior should be.

BeaconSage:
## Summary

Closes #1512.

This PR adds an opt-in auto-open review session feature for flashcards that are ready to learn. When enabled, the plugin periodically syncs, checks for new or due cards, and opens the existing review UI directly into a deck/card flow.

## What changed

- Added auto-open settings under Flashcard Review:
  - enable/disable automatic review sessions
  - check interval in minutes
  - optional startup check
  - configurable attention cues: notice, short sound, and dock bounce
- Reused the existing review queue, deck selection, review sequencer, and card UI.
- Added a dedicated auto-open path that prefers the last reviewed deck and falls back to the first deck with cards ready.
- Added foreground focus helpers for desktop review sessions.
- Avoided triggering while SR is already open, while the user is editing text, or while sync is locked.
- Added settings migration coverage and deck-selection tests.

## User-facing behavior

The feature is disabled by default. Users who enable it can choose how intrusive the reminder should be by turning notice, sound, and dock bounce cues on or off. The startup check is also disabled by default.

## Validation

- pnpm exec tsc --noEmit
- pnpm exec jest --runInBand
- pnpm lint
- pnpm build

## Notes

The implementation does not add a separate scheduling algorithm or review flow. It only automates opening the existing review session when the existing review queue contains cards ready to learn.


KyleKlus (changes requested):
Hi, I had a very broad look at this and I have a bit of feedback on this issue 😇:

I find the implemented usecase a bit to niche for the general user base, if we just add this as an auto open feature, as I think it is missing a proper integration into the review workflow.

So how about we build this into a proper reminder system 😊. 
I think first it would be great to have a reminders section on the schedule settings page, where one can enable (via toggle) getting reminders/notifications at customizable intervals (via some kind of date time selector). Then there should be a text area, where one can put a customizable text, which is displayed in the notification. Below that I would have another toggle, which enables your already implemented auto open feature, which opens the review at the same time as the reminder.

I think separating the auto open from the notification would make it less of a one of a kind use case, but a integrated reminder system with a few options, which one can expand upon in the future 😇.

What do you think of this😊?

KyleKlus (changes requested):
Then, if you agree to develop this further I would have another few suggestions😇:

- The code coverage tests need to be successful to be able to merge your code, which means that you have to cover the missing lines with some tests
- I would find it awesome if you also documented your code better with inline comments &  js doc comments [JSDoc Reference](https://medium.com/@uomroshan/a-comprehensive-guide-to-jsdoc-comments-in-javascript-ed14df2351ef). This makes maintaining/extending the code far easier, when one has such a big project


The back end code is undergoing quite a dramatic refactor right now in #1471, so it might be good to wait with any further implementation of this until it is merged, as I think it would be quite hard to fix the merge issues of merging yours first😅

BeaconSage:
Thanks, I agree with this direction.

After thinking about it, framing this as a general review reminder system makes more sense than shipping auto-open as a standalone feature. I like the proposed structure:

- a reminders section in scheduling settings
- a main toggle for reminders
- configurable reminder interval
- customizable notification text
- a separate toggle to auto-open the review session when a reminder fires

The current PR solved part of the underlying mechanics, but I agree the product shape should be broader and easier to extend.

To give a bit more context on the original use case: during my daily AI coding work, there are many short waiting periods where I am waiting for generation, tests, builds, or reviews. I want to use those small gaps to review knowledge with spaced repetition. Realistically, though, after work I often do not have enough energy or discipline left to open Obsidian and study deliberately.

That is why I originally leaned toward a more forceful “auto-open the card now” interaction. I had also considered building a reminder system, but what I personally wanted most was the speed of having a question card appear immediately so I could answer it with almost no friction. If I am too busy, I can close it. If I only have a tiny gap, I can answer one card and close it. For me, spaced repetition is important enough that I am willing to accept some interruption to my workflow if it helps me actually learn and review consistently.

In other words, I trust a tool that puts the card directly in front of me more than I trust my own discipline to remember to study later. That is the motivation behind the current implementation. But I agree that making this part of a broader reminder system, with auto-open as an optional behavior, would make it more useful for the general user base.

Given the overlap with the large refactor in #1471, I think it would be cleaner to pause major changes here and rework this after #1471 lands, either by rebasing this branch or opening a fresh PR for the reminder system.

KyleKlus:
Hi, the rework is now merged into everything, so if you like you can continue with this feature 😇

BeaconSage:
Hi, thanks for the feedback, and thanks for confirming that `#1471` has landed.

I agree that this should be reshaped from a standalone auto-open feature into a broader review reminder system, with auto-open kept as an optional behavior when a reminder fires.

I’ll continue the rework in this PR with that direction in mind. The next changes will focus on:

- reframing the feature around review reminders
- moving the settings to a reminder-oriented structure
- keeping the existing review queue and review UI instead of adding a separate flow
- preserving the existing safeguards around editing state, open SR state, and sync lock
- updating tests and settings migration coverage
- adding targeted inline/JSDoc comments where they help maintainability

KyleKlus (commented):
I've had a look at the code now and all looks nicely done😊.
Now I'll just have a look at the feature in obsidian itself and then this PR is ready to merge😇

KyleKlus:
Alright, everything works nicely. Thank you for your contribution 😊

PR #1425, opened 2026-03-23
feat: add FSRS scheduling for flashcards

kiankyars:
## Summary
- add FSRS as a flashcard scheduling option using `ts-fsrs`
- keep whole-note scheduling on the existing OSR path
- support FSRS schedule persistence in card comments, lazy import from legacy SM-2 card data, and short-term requeue/wait behavior for learning/relearning steps
- add the desired retention setting and targeted test coverage

## Notes
- uses the upstream `ts-fsrs` implementation rather than a custom FSRS scheduler
- preserves existing note review frontmatter behavior
- FSRS flashcard due checks use the current time instead of start-of-day comparisons
- includes a best-effort fallback path when switching back to OSR by preserving `interval` and projected `latestEase`, but it is not a lossless reconstruction of the original pre-FSRS schedule

## Validation
- `pnpm exec tsc --noEmit`
- `pnpm exec eslint src tests`
- `TZ=UTC pnpm test -- --runInBand`

## Related
- part of #1397

Closes #748
Closes #999
Closes #516

PR #1537, opened 2026-05-21
docs: add FSRS to README for discoverability

kiankyars:
This updates the README to explicitly mention the FSRS algorithm, addressing feedback from @freedomforwhat in https://github.com/st3v3nmw/obsidian-spaced-repetition/pull/1425. The user noted that without this, the plugin is not discoverable when searching for 'fsrs' in the Obsidian plugin gallery.


kiankyars:
Closing this because of the comment that @KyleKlus is waiting to update the review until the major release, which makes sense.

PR #1575, opened 2026-06-11
fix: Preserve FSRS sibling schedule placeholders

### LINKED ISSUE #1554: [Bug]: FSRS: Buried Sibling Cards Not Surfacing as "New" After Bury Clears + Corrupted Difficulty Values
## Summary

Two related bugs occur when using FSRS with the "Bury sibling cards until the next day" setting enabled on two-way (`:::`) cards:

1. **Buried sibling cards (from NEW cards) do not appear in the review queue the next day** (or at least not as NEW), even though the bury list has correctly cleared, hence they probably get assigned some scheduling info and a revision count during the "bury" process.
2. **Some reviewed siblings are assigned a wildly out-of-range difficulty value** (way above 10) on cards that were created and reviewed entirely under FSRS — not migrated from the previous algorithm.

---

## Environment

- **Plugin:** obsidian-spaced-repetition
- **Plugin version:** latest (v1.15.1)
- **Algorithm:** FSRS (freshly switched from SM-2)
- **Setting:** "Bury sibling cards until the next day" — **enabled**
- **Card type:** Two-way / reversed (`:::`)
- **OS:** MacOS latest

---

### Observed behaviour

The buried reverse card has valid FSRS scheduling data written into the note comment, with a due date of the previous day and interval of 0. The bury list clears correctly overnight. Despite this, the card does not appear in the review queue as NEW.

### Relevant scheduling comment (reverse/buried sibling block)

```
!fsrs,2026-05-31T13:48:30.331Z,0,2.3065,2.11810397,1,1,0,1,2026-05-31T13:38:30.331Z
```

Decoded:

|Field|Value|Notes|
|---|---|---|
|Due|2026-05-31T13:48Z|Previous day — overdue|
|Interval|0|Same-day scheduling|
|Difficulty|2.3065|Normal range|
|Stability|2.118|~2 days|
|State|1|Learning|
|Reps|1|Reviewed once|
|Lapses|0|—|
|Last review|2026-05-31T13:38Z|Previous day|

### Hypothesis

The queue builder may not be correctly re-ingesting cards that are in **Learning state with interval=0** after they have been through a bury cycle. The bury mechanism writes a valid due date for the sibling at review time, but the queue construction logic on the following day appears to drop these cards.

---

### Observed behaviour

The forward card (the one actually reviewed, not buried) consistently receives a difficulty value above 10, and sometimes an exact same value (i.e. `18.76074875` was assigned to at least 3 cards) — far outside FSRS's valid difficulty range of 1–10. This has been observed identically across multiple cards, all created and reviewed entirely under FSRS with no prior SM-2 history.

### Relevant scheduling comment (full two-sibling comment)

```
<!--SR:!fsrs,2026-06-20T06:13:53.252Z,19,18.76074875,6.99791867,2,2,0,0,2026-06-01T06:13:53.252Z!fsrs,2026-05-31T13:47:14.565Z,0,2.3065,2.11810397,1,1,0,1,2026-05-31T13:37:14.565Z-->
```

Forward card decoded:

|Field|Value|Notes|
|---|---|---|
|Due|2026-06-20|19 days out|
|Interval|19|days|
|Difficulty|**18.76074875**|⚠️ Outside valid range (1–10)|
|Stability|6.998|~7 days|
|State|2|Review (graduated)|
|Reps|2|—|
|Lapses|0|—|
|Last review|2026-06-01|Today|

### Key observations

- The **same value (`18.76074875`) appears across multiple distinct cards**, suggesting it is a deterministic output of a specific code path rather than random corruption.
- The **reverse sibling of the same card has a normal difficulty value** (`2.3065`), despite being generated in the same session.
- The cards were **created and reviewed entirely under FSRS** — this is not a migration artefact from SM-2.
- The only structural difference between the two siblings at review time is that **the forward card was reviewed while the reverse was being buried simultaneously**.

### Hypothesis

The difficulty corruption is triggered specifically by the bury interaction during the **first review** of a two-way card. When FSRS writes the forward card's scheduling data while simultaneously burying the sibling, an incorrect value is passed into the difficulty calculation — possibly an unscaled intermediate, an index, or a value from the wrong field. The bury event appears to corrupt the difficulty of the card that _was_ reviewed, not the one that was skipped.

### To Reproduce

1. Enable FSRS and enable "Bury sibling cards until the next day" in settings.
2. Create a new two-way card using `:::` syntax (brand new, never reviewed under any algorithm).
3. Review the deck. The forward card is shown; the reverse card is buried.
4. Rate the forward card (e.g. Good or Easy).
5. The bury list correctly records the reverse sibling.
6. The next day: open Obsidian. Confirm the bury list has cleared.
7. **Expected:** The reverse sibling appears as due today as NEW (Learning state, interval 0, overdue).
8. **Actual:** The reverse sibling does not appear in the review queue as NEW (although it probably appears as already reviewed and dew, couldn't verify this due to my very long review queue) The overall card count in the deck is correct.

## Expected Behaviour

1. Buried sibling cards with a past due date and Learning state should appear in the review queue the day after the bury clears. Buried NEW never reviewed siblings should appear then next day as NEW and not be assigned scheduling info and a revision count.
2. FSRS difficulty values should always be initialized and updated within the valid range (approximately 1–10). Values such as `18.76074875` should never be produced for a card reviewed for the first time under FSRS.

### Screenshots

_No response_

### OS of your device

Mac OS

### Obsidian version

1.12.7

### Plugin version

1.15.1

### Installer version

1.12.4

## Additional Example Comments

**Card with asymmetric migration (forward on FSRS, reverse still on SM-2):**

```
<!--SR:!fsrs,2026-07-01T06:09:52.035Z,30,30.46452326,9.53567837,2,5,0,0,2026-06-01T06:09:52.035Z!2027-03-04,364,256-->
```

The reverse sibling here was never migrated to FSRS and carries an SM-2 due date of 2027-03-04 (364-day interval). Note the forward card also has an out-of-range difficulty (`30.46`). This particular card was migrated from SM-2 and reviewed once (one of the siblings) under FSRS.

### Config file

_No response_

kiankyars:
## Summary

Fixes #1554.

This PR addresses the FSRS/reversed-card sibling bugs described in the issue and follow-up comments:

- keeps placeholder schedule entries in multi-card HTML comments instead of dropping them during parsing
- fixes dummy `2000-01-01` schedule detection so placeholders parse as empty/new card slots
- fixes the random reversed-card pending requeue path where the back side can be shown first and sibling deletion can make the iterator's current card disappear

## Why

With FSRS plus "bury sibling cards" on reversed (`:::`) cards, review state is stored as one schedule entry per sibling. Placeholder entries like `!2000-01-01,1,250` represent unscheduled/new siblings. Dropping those placeholders collapsed the schedule array and shifted a later FSRS schedule onto the wrong sibling after reload.

Separately, when random order shows the reverse side first, `handlePendingRequeue` deleted sibling cards before deleting the current pending card. If the deleted sibling was earlier in the deck list, the current index could point at the wrong card or no card, causing the undefined-card error described in the issue comments and leaving review flow blocked.

## Tests

Added targeted unit coverage for:

- placeholder-only schedule comments parsing as empty card slots
- placeholder-before-FSRS schedule comments preserving sibling indexes
- placeholder-after-FSRS schedule comments preserving sibling indexes
- placeholder/FSRS/placeholder comments preserving a middle sibling schedule
- reversed `:::` cards keeping the FSRS schedule on the reviewed reverse sibling
- reversed `:::` cards keeping the unreviewed reverse sibling new when the placeholder follows FSRS
- pending FSRS requeue with `burySiblingCards=true` when random order shows the reverse side first

Validation run locally:

- `pnpm exec tsc --noEmit --skipLibCheck`
- `pnpm lint` (passes with existing warnings only)
- `pnpm jest --runInBand --coverage=false` (42 suites, 366 tests)
- `git diff --check`

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -59,13 +59,13 @@ importers:
         version: 9.2.1
       esbuild:
         specifier: ^0.28.0
-        version: 0.28.0
+        version: 0.28.1
       eslint:
         specifier: ^10.4.1
         version: 10.4.1
       eslint-plugin-obsidianmd:
         specifier: ^0.3.0
-        version: 0.3.0(@eslint/js@10.0.1(eslint@10.4.1))(@eslint/json@0.14.0)(@typescript-eslint/parser@8.60.1(eslint@10.4.1)(typescript@5.9.3))(eslint@10.4.1)(obsidian@1.13.0(@codemirror/state@6.6.0)(@codemirror/view@6.43.0))(typescript-eslint@8.60.1(eslint@10.4.1)(typescript@5.9.3))
+        version: 0.3.0(@eslint/js@10.0.1(eslint@10.4.1))(@eslint/json@0.14.0)(@typescript-eslint/parser@8.60.1(eslint@10.4.1)(typescript@5.9.3))(eslint@10.4.1)(obsidian@1.13.0(@codemirror/state@6.6.0)(@codemirror/view@6.38.6))(typescript-eslint@8.60.1(eslint@10.4.1)(typescript@5.9.3))
       eslint-plugin-simple-import-sort:
         specifier: ^12.1.1
         version: 12.1.1(eslint@10.4.1)
@@ -86,7 +86,7 @@ importers:
         version: 2.30.1
       obsidian:
         specifier: ^1.13.0
-        version: 1.13.0(@codemirror/state@6.6.0)(@codemirror/view@6.43.0)
+        version: 1.13.0(@codemirror/state@6.6.0)(@codemirror/view@6.38.6)
       pre-commit:
         specifier: ^1.2.2
         version: 1.2.2
@@ -95,7 +95,7 @@ importers:
         version: 3.8.3
       ts-jest:
         specifier: ^29.4.11
-        version: 29.4.11(@babel/core@7.29.0)(@jest/transform@30.4.1)(@jest/types@30.4.1)(babel-jest@30.4.1(@babel/core@7.29.0))(esbuild@0.28.0)(jest-util@30.4.1)(jest@30.4.2(@types/node@25.9.1)(ts-node@10.9.2(@types/node@25.9.1)(typescript@5.9.3)))(typescript@5.9.3)
+        version: 29.4.11(@babel/core@7.29.0)(@jest/transform@30.4.1)(@jest/types@30.4.1)(babel-jest@30.4.1(@babel/core@7.29.0))(esbuild@0.28.1)(jest-util@30.4.1)(jest@30.4.2(@types/node@25.9.1)(ts-node@10.9.2(@types/node@25.9.1)(typescript@5.9.3)))(typescript@5.9.3)
       tslib:
         specifier: ^2.8.1
         version: 2.8.1
@@ -282,8 +282,8 @@ packages:
   '@codemirror/state@6.6.0':
     resolution: {integrity: sha512-4nbvra5R5EtiCzr9BTHiTLc+MLXK2QGiAVYMyi8PkQd3SR+6ixar/Q/01Fa21TBIDOZXgeWV4WppsQolSreAPQ==}
 
-  '@codemirror/view@6.43.0':
-    resolution: {integrity: sha512-V7ZCLQO3Jus9hzh2jVCCPW3mO4IBMr43O37PqSUYautJSnnJF41YlgLw21x0fLJTYvJ+Vkm6Gp+qKGH9pltgXA==}
+  '@codemirror/view@6.38.6':
+    resolution: {integrity: sha512-qiS0z1bKs5WOvHIAC0Cybmv4AJSkAXgX5aD6Mqd2epSLlVJsQl8NG23jCVouIgkh4All/mrbdsf2UOLFnJw0tw==}
 
   '@cspotcode/source-map-support@0.8.1':
     resolution: {integrity: sha512-IchNf6dN4tHoMFIn/7OE8LWZ19Y6q/67Bmf6vnGREv8RSbBVb9LPJxEcnwrcwX6ixSvaiGoomAUvu4YSxXrVgw==}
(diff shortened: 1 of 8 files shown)


KyleKlus (approved):

KyleKlus:
Hi, 
First off thank you so much for working on these fixes😊.

Everything looks fine except one bug when all cards are reviewed and one wants to open the decks list via the select a deck to review command. The bug doesn't happen, when the bury feature isn't turned on and when the burylist has some content in it. Also I think the bug doesnt appear, once there are again cards which are due, this only happens when there a no due cards at all🤔.

To reproduce:

1. Create a few reversible cards in a note
2. Turn on the bury feature
3. Make sure the bury list in data.json is empty
4. Review all the cards
5. After that open the decks list again via "Review flashcards from all notes"

Expected:

The deck list opens, all decks are shown, but marked as not reviewable (greyed out)

Current behavior (bug):

A blank sr view opens and the following error appears ->
````
Failed to open view TypeError: Cannot read properties of null (reading 'getTopicPath')
    at ContentManager._getNewSessionData (plugin:obsidian-spaced-repetition:11865:40)
    at ContentManager._reviewDeck (plugin:obsidian-spaced-repetition:11798:29)
    at ContentManager.open (plugin:obsidian-spaced-repetition:11780:18)
    at async SRTabView.onOpen (plugin:obsidian-spaced-repetition:12126:5)
````

With this being the code that it points to -> 
````
_getNewSessionData(deck) {
    if (this.reviewSequencer === null) return null;
    const deckStats = this.reviewSequencer.getDeckStats(deck.getTopicPath());
    const totalCardsInSession = deckStats.cardsInQueueCount;
    const totalDecksInSession = deckStats.decksInQueueOfThisDeckCount;
    const currentCardState = 0 /* Front */;
    const currentDeckStats = this.reviewSequencer.getDeckStats(
      this.reviewSequencer.currentDeck.getTopicPath()
    );
    const chosenDeckStats = this.reviewSequencer.getDeckStats(deck.getTopicPath()); <---------- Here
    return {
      cardData: {
````

KyleKlus (approved):

KyleKlus:
I found out what caused the bug😇. It was one of my own mistakes when retrieving info about if there are some due cards😅

Issue #1234, opened 2025-02-25
[FEAT]: Ability to store data in a seperate file.


Vishvesh1542:
### Is your feature request related to a problem? Please describe.

It would be very helpful if we could store the data regarding the difficulty of a flashcard etc in a separate file and link to it using IDs maybe. It would be really helpful for people sharing flashcards and not being able to use the feature because each of them has different data and it is really hard to separate them 

### Describe the solution you'd like

Just adding an option to save the data in a separate file and linking it to the note using IDs. It would also be better if it didn't create new IDs when it detects one ( For example friend makes flashcard and his plugin creates an ID. And when I use the same flashcard, it detects the ID and uses that too.

KyleKlus:
Hi, I am closing this as a duplicate of #1538, as that issue keeps track of the whole progress on this feature😇

Issue #957, opened 2024-05-17
[FEAT] More synchronization-friendly settings


mehanoid:
**Is your feature request related to a problem? Please describe.**
In the plugin settings there is a `buryDate` property in the data.json file. The plugin writes new values for this property every day, and this may cause frequent synchronization conflicts, for example, when using git for synchronization.

**Describe the solution you'd like**
Maybe it would be better to put such volatile properties in a separate file so that you can add it to gitignore.

4Source:
I would agree with this. Maybe remove the ``buryDate``, ``buryList`` completely from the settings and figure it out with the stored information at the end of a card. 

For others to Test this: 
[data.json](https://github.com/st3v3nmw/obsidian-spaced-repetition/files/15415632/data.json)
[Bury.sibling.cards.md](https://github.com/st3v3nmw/obsidian-spaced-repetition/files/15415661/Bury.sibling.cards.md)

mehanoid:
@ronzulu multiple of my own devices: laptops with obsidian-git plugin and android phone with termux and tasker.
BTW I doubt it makes sense to synchronise spaced-repetition with other people 😃

4Source:
> BTW I doubt it makes sense to synchronise spaced-repetition with other people 😃

I think if it is only about the settings it could be but then there can't be any flashcards related data in the data.json

ronzulu:
> @ronzulu multiple of my own devices: laptops with obsidian-git plugin and android phone with termux and tasker. BTW I doubt it makes sense to synchronise spaced-repetition with other people 😃

Thanks for the info @mehanoid .

The reason for the question is that I remember someone saying that they were studying medicine and wanted to share notes and flashcards with their fellow students. The issue for them being that sharing files is greatly hampered because of the scheduling info embedded in the actual files.

They were after a mechanism to separate out the note/card content from their own personal scheduling info.

PR #1579, opened 2026-06-14
feat: store scheduling data in a single vault file (StorageType.SR_MD)

### LINKED ISSUE #1538: [FEAT]: Store schedule data outside of notes (Help wanted)
Hi, this is a issue which collects the state of this feature and it documents what has been decided on how to implement this.
This is so that who ever wants to implement it has a clear picture on how to do it.

In general we have agreed that this kind of data store should work as follows:

## Storing data

- In the vault in a customizable location (by default in "/Spaced Repetition"), as the obsidian sync service doesn't sync random files in the plugin folder by default and because it is hard to create/update files in the plugin folder on mobile
- The schedule data will be stored in a file named "schedule-data.sr.md" which is located in the "Spaced Repetition" folder
- The data is stored in json format, but still in a markdown file, so that the "obsidian sync service" syncs it by default

## Linking data

- Each reviewable card gets a block id like this ("sr-id" is used to make them easy to find)-> ^sr-id-[generated short uuid goes here]
- Each reviewable note gets a sr-id in the frontmatter like it is already sort of done
- Those ids link then to the schedule data files

## More resources

There has already been some progress on this which can be found in the data store folder in a zip file called "WIP_DataStores". So feel free to have a look at this to maybe reuse some of the code.

In there are files which already kind of implement that kind of data store, but it never really worked out.
Maybe you can use them as a starting point for your own data store. Also have a look at this PR for a more detailed infos on this: #1471

### List of issues requesting it

- #1234
- #1465
- #162
- #957

Lorite:
## Summary

Adds an **opt-in** data store (`StorageType.SR_MD`) that keeps **all** scheduling data in a single markdown file in the vault — `<configurable folder>/schedule-data.sr.md` (default folder `Spaced Repetition`) — as JSON inside a fenced code block, instead of inline `<!--SR:...-->` comments and `sr-*` frontmatter.

This implements the design agreed in #1538 and is a fresh re-implementation against current `main` (it supersedes #1471, whose feature was shelved during the backend refactor). It also addresses the long-standing requests in #1465, #1234, #162 and #957.

The default store remains `NOTES`; nothing changes unless the user switches in **Settings → Data → Storage of Scheduling Data**.

### Why a file in the vault (not plugin data)
Per the discussion in #1538: a vault file is synced by Obsidian Sync by default (plugin-folder files are not) and is writable on mobile.

### Linking model (as agreed in #1538)
- Each reviewable **card** gets an Obsidian block id `^sr-id-<short-uuid>` written next to it; its schedule lives in the file keyed by that id.
- Each reviewable **note** gets an `sr-id` frontmatter field; its note-level schedule lives in the file keyed by that id.
- Both survive renames/moves, inside or outside Obsidian.

## Design notes

- **Synchronous read path (no parser async ripple).** The schedule file is preloaded into memory once before each vault scan (`IDataStore.preload`), so `createSchedule`/`removeScheduleInfo` stay synchronous and the question parser is untouched. This deliberately avoids the async-everywhere refactor.
- **Reuses existing serialization.** Stored values are the same inline schedule strings that go inside `<!--SR:...-->`, so `CommentParser` + the existing note algorithm handle both **OSR and FSRS** with no duplicated logic.
- **Bidirectional migration** (`NOTES ⇄ SR_MD`) via a two-phase *read-with-one-store / write-with-the-other* approach that reuses the parser and the stores' own write paths instead of bespoke text surgery. Confirmation modal included.
- **Null/parse-tolerant repository.** A missing or corrupt schedule file yields empty data and never throws, so it can't break loading the vault.
- **Bug fix:** anchored the jest `moduleNameMapper` to `^src/` so third-party modules whose internal requires contain `src/` (e.g. `short-uuid`'s `./src/converter`) resolve correctly.

## Files
New store under `src/data/data-store/sr-md-data-store/` (repository, store, algorithm, file modifier, migrator); wiring in `data-manager.ts` / `core.ts`; settings UI in `data-page.tsx`; new locale keys; unit tests for the repository, the store (incl. a parser integration test), and the algorithm.

## Testing
- `pnpm test` — **383 passing**, 44 suites, coverage thresholds met.
- `pnpm run build` — clean.
- Manually smoke-tested in a real vault: switching to SR_MD creates `schedule-data.sr.md` with JSON, reviews persist across reloads, notes gain `^sr-id-` block ids + `sr-id` frontmatter and **no** `<!--SR:-->`, and switching back to NOTES restores inline comments and clears the file.

## Notes for maintainers
- **Base branch:** opened against `main` because the `next-*` release branches don't currently exist on the repo — happy to retarget to `next-minor-version`/`next-major-version` per your release flow.
- Opened as a **draft** to coordinate on #1538 before final review. Questions: is the `schedule-data.sr.md` filename/JSON-in-markdown format what you want long-term? Should note-level schedules move into the file too (done here) or stay in frontmatter for v1?

Refs #1538, #1465, #1234, #162, #957


KyleKlus:
Hi, first off; thank you for having another go at this😊, especially after I botched your last PR 😅.

So for now it would be fine to keep them both in the same file. I think if later requested then one could move the note schedules into a separate file😇.

My plan is to unify the implementation of note and card scheduling at some point (It is wild to me that there are two separate implementations of basically the same thing😅) and I think it would be nice if the data stores already treat them the same.

Lorite:
Hi, and sorry for the long silence on this, life got in the way.

That works for me. The current implementation already keeps cards and notes in the same file, with two maps keyed by id (`cards` by block id, `notes` by `sr-id`), so splitting them later would just be a change to the repository layer.

Unifying note and card scheduling sounds good to me, and I agree the data stores should treat them the same. Right now `IDataStore` still has separate `readNoteSchedule`/`writeNoteSchedule` and `createSchedule`/`writeSchedule` pairs because that mirrors the existing split. If you want, I can converge them in this PR, or keep this one as-is and do the unification separately once you have decided on the shape.

I will also push a fix for the patch-coverage failure.

Issue #1580, opened 2026-06-14
[Perf] Flashcard/note review re-reads & re-parses the entire vault on every invocation (no caching / incremental sync)


Lorite:
## Summary

Every flashcard review (and note review) triggers a **full re-read and re-parse of the entire vault**. `sync()` rebuilds the complete in-memory model from scratch each time it runs, and for every note that has a flashcard tag/folder it does an **uncached disk read + full markdown parse** to extract its cards. There is no caching and no incremental update (no `vault`/`metadataCache` change listeners), so the cost is paid in full on every invocation and scales with the size of the whole vault rather than with the work actually being done.

On large vaults this makes opening the review modal take many seconds.

## Where it happens (line refs against `main` @ 22fbea0)

The flashcard review path calls `sync()` on every invocation:

- [`review-queue-loader.ts#L57`](https://github.com/st3v3nmw/obsidian-spaced-repetition/blob/22fbea0a71ebd39fe29e0bc2d571aea4c42e10e4/src/ui/review-queue-loader.ts#L57) → `await this.plugin.dataManager.sync();`

`sync()` → `loadVault()` iterates **every** markdown file in the vault and processes each one:

- [`data-manager.ts#L130`](https://github.com/st3v3nmw/obsidian-spaced-repetition/blob/22fbea0a71ebd39fe29e0bc2d571aea4c42e10e4/src/data/data-manager.ts#L130) → `const notes = this.plugin.app.vault.getMarkdownFiles();`
- [`data-manager.ts#L142`](https://github.com/st3v3nmw/obsidian-spaced-repetition/blob/22fbea0a71ebd39fe29e0bc2d571aea4c42e10e4/src/data/data-manager.ts#L142) → `await this.osrCore.processFile(file);`

For each note with a flashcard topic path, `processFile()` loads and fully parses the note body:

- [`core.ts#L219`](https://github.com/st3v3nmw/obsidian-spaced-repetition/blob/22fbea0a71ebd39fe29e0bc2d571aea4c42e10e4/src/data/core.ts#L219) → `note = await this.loadNote(noteFile, topicPath);` → `NoteFileLoader` → `NoteQuestionParser`

…and the actual read is **uncached** (`vault.read`, not `vault.cachedRead`):

[...]

PR #1587, opened 2026-06-24
perf: use cachedRead for flashcard parsing (avoid a disk read per note on every review)

Lorite:
## What & why

`OsrCore.loadVault()` parses every flashcard note on every sync — and a sync runs on every flashcard/note review (`review-queue-loader.ts`), every "Review a note" command, and every reminder tick. Each note parse does an **uncached** `vault.read()` of the whole file, so a single review re-reads every flashcard note from disk. On large vaults — and especially on synced / mobile vaults — this makes opening review take seconds.

This is the smallest, independent slice of #1580 (the disk-I/O cost), and is safe to land on its own.

## Changes

- Add `ISRFile.cachedRead()` (backed by Obsidian's `Vault.cachedRead`) and use it on the **read-only** parse path (`NoteQuestionParser`). Obsidian serves unchanged files from its in-memory cache, turning a disk read per note into a memory read on every sync.
- Deliberately keep `read()` on the read-modify-write paths (block-id assignment, schedule writes, deletes, `writeNoteFile`) — a stale cached copy there could clobber a recent edit. `cachedRead` is documented as read-only-only.
- De-duplicate the `readNoteSchedule` call in `OsrCore.processFile` (it was read twice per file).

## Scope / follow-ups

This intentionally only removes the **disk read**; the markdown re-parse still runs each sync. The larger follow-ups discussed in #1580 — caching parsed notes to skip the re-parse, and event-driven incremental sync to skip unchanged syncs entirely — build on top of this and will come as separate PRs.

## Testing

- `tsc -noEmit -skipLibCheck` — clean
- full unit suite — passing
- production `esbuild` build — OK

Refs #1580


Lorite:
Friendly ping on this one. It is small and self-contained (uses `Vault.cachedRead` on the read-only parse path so a sync stops doing a disk read per flashcard note), CI is green and it still merges cleanly. Happy to rebase or split it further if that helps.

PR #1588, opened 2026-06-27
feat: Add interactive Multiple Choice Question support

mazzoccantelorenzo:
Native support for interactive Multiple Choice Questions.
Users can now use standard markdown checklists on the front of a card to create clickable MCQ options.

Simply use a multi-line separator (`?` or `??`) and place a checklist on the front side, marking the correct answer(s) with `[x]`.

This is a file i've used to test it out.
[MCQ-Test-Deck.md](https://github.com/user-attachments/files/29421580/MCQ-Test-Deck.md)

And this is a show case of what i've been using. I think it's very easy to use, also helps me play more with cards.

https://github.com/user-attachments/assets/a143b924-f260-4122-809a-1a68eab4e267

   
I've made sure the implementation is lightweight and purely CSS/UI driven so it doesn't break any existing scheduling logic. Let me know if you need any changes ( probably yes ahah) !


mazzoccantelorenzo:
Do we merge?

Issue #282, opened 2021-08-29
Help Translate the Plugin - Multilingual support


st3v3nmw:
Please check [this](https://github.com/st3v3nmw/obsidian-spaced-repetition/wiki/Help-Translate) for more details.

TL;DR
1. Fork this repository,
2. Copy the entries from `src/lang/locale/en.ts` to the proper file in `locales` (i.e. `fr.ts` for French, or `sw.ts` for Swahili),
3. Translate
4. Then open a pull request.
5. & a thank you for your time, much appreciated!

GollyTicker:
Since it's just 130 keys to translate, I started the translations for the german `de.ts` locale.

Here is the file: https://github.com/GollyTicker/obsidian-spaced-repetition/blob/feature/locale-de/src/lang/locale/de.ts

I'll open the pull request once I'm finished.

st3v3nmw:
Hi @GollyTicker, the German translation is now available in the new release (v1.7.0).

Japanese translation too (thanks to @yo-goto).

christopherAI2022:
Start one for Spanish too please!

ALizarazoTellez:
Hello, I have sent the translations of the Spanish language. I apologize if there were any errors, I have no experience doing translations and my English is not very fluent; I am a native Spanish speaker.

Please @christopherAI2022, check my translations.

The Pull Request is the number #667.

Hola, he enviado las traducciones del idioma español. Me disculpo si llegase a haber algún error, no tengo experiencia haciendo traducciones y mi inglés no es muy fluido; soy hablante nativo del español.

Por favor @christopherAI2022, revisa mis traducciones.

El Pull Request es el número #667.

PR #1596, opened 2026-07-10
feat: complete the Spanish (es) translation

ymxlx:
## What

Completes the Spanish (`es`) translation, working towards #282.

`es.ts` spreads `...en` and then overrides individual keys, so any key it doesn't override silently renders in English. I found that **exactly 100 keys** were falling back to English — mostly the newer settings: the review-reminder options, the scheduling-data storage/migration flows, the delete-scheduling-data actions, the status-bar items, the FSRS switch confirmation, reset settings, and a few flashcard-modal strings (`Delete Card`, `Jump to card`, `Cram Mode`, …).

This PR translates all 100. After the change, `es.ts` overrides every key defined in `en.ts` (319/319).

## How

- Added the 100 missing keys, grouped by source file with comments matching the existing structure.
- Matched the terminology already used in `es.ts` for consistency (`Mazos`, `Tarjetas`, `datos de programación`, `Facilidad`, …).
- Preserved every interpolation placeholder (`${difficulty}`, `${docsUrl}`, `${totalCardsCount}`, …) and inline HTML (`<a href="…">`) exactly.

## Verification

- `tsc --noEmit` passes with 0 errors.
- Verified programmatically that the set of added keys is exactly the set that was falling back, with no duplicates.

## Scope note

A handful of keys were already *present* in `es.ts` but left with English values (e.g. `NOTES`, `NOTE`, `VIEW_CARD_INFO`, some `GROUP_*` headers). I left those out of this PR to keep the diff focused on the fall-through keys; happy to follow up with those in a separate PR if you'd like.


st3v3nmw (approved):
Thank you.

PR #1606, opened 2026-07-22
fix: refresh flashcard UI immediately after edit Save

kiankyars:
## Summary

- After pressing **Save** in the pencil edit modal, the markdown note was updated correctly but the open flashcard review view could still show the previous front/answer text.
- Root cause: the review session was not reliably re-synced from the sequencer after `updateCurrentQuestionTextAndCards`, redraw depended on `uiManager.uiState` (which is `EditModal` during the edit flow and can drift from the displayed card side), and a card-count mismatch aborted all in-memory `front`/`back` updates even though the file had already been written.
- This PR always refreshes session card/question/note from the sequencer, redraws from the displayed card side (`CardFront` / `CardBack`), updates overlapping in-memory faces after every successful write, and hardens null-safe drawing / toolbar stats.

## Changes

- `content-manager.tsx`: `_doEditQuestionText` now restores via `_refreshCurrentCardView`, keyed off session card state rather than UI state alone.
- `flashcard-review-sequencer.ts`: always update overlapping card faces after write; warn if sibling count changed instead of leaving stale faces.
- `card-container.tsx` / `toolbar.tsx`: null-safe current card and deck stats during redraw.
- Tests: assert in-memory `front`/`back` update after edit for single-line and multiline cards.

## Test plan

- [x] `jest` `updateCurrentQuestionTextAndCards` + Sequences suites
- [x] `tsc -noEmit -skipLibCheck`
- [ ] Manual: open flashcard review → pencil edit front and/or answer → Save → confirm the open card (and Show Answer side) updates immediately without leaving/reopening review
- [ ] Manual: edit while answer is already shown → Save → confirm both sides refresh
- [ ] Manual: cloze edit that changes deletion count → Save → current card text updates; notice about sibling re-parse is acceptable

### DIFF
--- a/src/scheduling/flashcard-review-sequencer.ts
+++ b/src/scheduling/flashcard-review-sequencer.ts
@@ -440,16 +440,24 @@ export class FlashcardReviewSequencer implements IFlashcardReviewSequencer {
 
         await this.currentQuestion.writeQuestion(this.settings);
 
-        if (cardFrontBackList.length !== question.cards.length) {
-            console.warn("SR: Cards count does not match question text. Skipping redraw.");
-            new Notice("Cards count does not match cards from question text. Skipping redraw.");
-            return;
-        }
-        question.cards.forEach((card, i) => {
+        // Always refresh in-memory faces for the overlapping cards so the review UI can redraw
+        // after Save. When the number of generated cards changes (e.g. cloze deletions), we
+        // still update as many as possible and notify the user that sibling count changed.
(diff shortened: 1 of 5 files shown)


kiankyars:
Closing this PR.

The reported issue (card not refreshing after edit Save) is fixed by upgrading to the latest release (v1.15.4+ includes the redraw-after-edit work). The additional hardening in this branch is not needed for that bug, so leaving it unmerged.

Thanks for confirming after the upgrade.

Issue #1628, opened 2026-08-19
[Bug]: The top right close button is displayed twice with a small offset


paulclrt:
### Describe the bug

On every window, the top right close button is displayed twice.
Clicking any of the two works perfectly fine. This is just a tiny UI problem.
See attached image for a better visual explaination.

### To Reproduce

1. Open obsidian
2. Install the plugin
3. Create a flashcard
4. Open the review flashcard popup by clicking the sidebar button
5. Notice the double close button on the top right.

### Expected behavior

There should be only one close button.

### Screenshots

<img width="1170" height="164" alt="Image" src="https://github.com/user-attachments/assets/72fdc7fb-67bc-4c6b-99e2-f1dd553ffe2f" />

<img width="461" height="96" alt="Image" src="https://github.com/user-attachments/assets/829e3e78-6858-4555-b590-7ff02c719a25" />

### OS of your device

Fedora - kernel 7.1.5-201.fc44.x86_64

### Obsidian version

1.13.7

### Plugin version

1.15.4

paulclrt:
I just noticed this is also a problem on my mobile device (IOS 26.6.1)

![image](https://github.com/user-attachments/assets/04393c0c-42e0-4758-bfbd-4ce7ae3f1fed)

rongeur-deviant:
i'm having the exact same issue here! also obsidian version 1.13.7 and plugin version 1.15.4, my os is windows 11 famille insider preview - version 25H2 - build 26220.9022

<img width="960" height="564" alt="Image" src="https://github.com/user-attachments/assets/219743f3-9ec6-4ca3-81b8-2e4d367bd123" />

update : i managed to fix the issue (cf next picture) by uploading a css snippet (settings > appearance > css snippets) with the following code :

#sr-modal-view > .modal-header-button {
    display: none !important;
}

<img width="960" height="564" alt="Image" src="https://github.com/user-attachments/assets/ec2ba7de-e249-4d69-8cf1-2c9670094bf6" />

Issue #1616, opened 2026-08-05
[Bug]: There are two close icons in the decks dialog


arniu:
### Describe the bug

<img width="657" height="517" alt="Image" src="https://github.com/user-attachments/assets/8686a44f-b85a-4dbc-ba7e-76c4db08f176" />

### To Reproduce

...

### Expected behavior

ONLY one

### OS of your device

MacOS 15.7.8

### Obsidian version

v1.13.4

### Plugin version

v1.15.4

### Installer version

5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) obsidian/1.13.4 Chrome/150.0.7871.114 Electron/43.1.1 Safari/537.36

PR #1635, opened 2026-08-27
fix: Hiding second close button by targeting their css classes

paulclrt:
I created an issue about this bug #1628 and just noticed it was also mentioned in another one #1616.
This fix simply uses the css class difference to hide the second button.

The fix is working on desktop and mobile:
Mobile:
image
image
Desktop
image


paulclrt:
My bad wrong branch. Really sorry about that

